PR #105, opened 2018-10-22
Make sleuth an optional dependency

ST-DDT:
I'm not 100% sure about this change, but at least for dependent maven projects sleuth was a required dependency. This PR changes this to make sleuth an optional dependency.

Feel free to request changes or discuss changes.


yidongnan:
In fact, CI uses the ``gradle publish`` deploy to the maven central repository, so the generated pom file don't include sleuth, but if there is a problem by using ``maven deploy``

PR #109, opened 2018-10-23
 Add the proto reflection service. Fixes #88

ST-DDT:
Fixes #88 

Added the proto reflection service to the services, and also added properties to turn the reflection service and the health service off.

I hope it is okay that I also added some javadocs and run the code formatter.
If not I can move that to a separate PR.


grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/springboot/autoconfigure/grpc/server/GrpcServerFactory.java
```diff
@@ -4,21 +4,63 @@
 import net.devh.springboot.autoconfigure.grpc.server.codec.GrpcCodecDefinition;
 
 /**
- * User: Michael
- * Email: yidongnan@gmail.com
- * Date: 5/17/16
+ * A factory that can be used to create grpc servers.
+ *
+ * @author Michael (yidongnan@gmail.com)
+ * @sice 5/17/16
```
yidongnan:
```suggestion
 * @since 5/17/16
```
It looks like a spelling mistake here.

ST-DDT:
Nice!

PR #118, opened 2018-10-29
Add missing javadoc

ST-DDT:
This adds javadocs to the rest of the classes/methods that were omitted in previous runs.
The only exception to that goal is the client lib because I'm not 100% sure how those components work and thus the javadocs would end up too inaccurate.

For the formatting I used the new layout/formatter, but I guess we still have to make a separate format commit after we add the automated tooling for it.


ST-DDT:
You were faster...
I guess we should first merge the formatting PR (because it is bigger)
And then I fix the merge conflicts.

ST-DDT:
Updated

PR #122, opened 2018-11-07
Add support for Spring-Security/Authentication

ST-DDT:
Related: #49 
Fixes: #50 , #69 

* Adds support for client authentication via certificates or basicAuth
* Adds support for server side authentication via certificates or basicAuth
* Adds support for service side authorization checks via Spring-Security's annotations or manual configuration
* Documented these features in the README

PR #126, opened 2018-11-12
Fixed multithreading issue in security layer

ST-DDT:
Improves #122 

In gRPC each callback can be called from a different thread, thus the current setup (which assumes that after the `startCall` method everything is called by the same thread) is wrong.

This PR changes the authenticating interceptor to work similar to grpc's Contexts which also use `ThreadLocal`s.

See also:
* [this question on SO](https://stackoverflow.com/questions/34905499/grpc-java-pass-data-from-server-interceptor-to-rpc-service-call)
* [grpc-Contexts](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/Contexts.java#L44)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ bin/
 .project
 .settings/
 .factorypath
+.checkstyle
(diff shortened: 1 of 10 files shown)


yidongnan:
I also passed the test many times, good job 👍

PR #131, opened 2018-11-13
Improve tracing

ST-DDT:
Remove bean dependency on auto config.
Fix conflict/creating `GrpcTracing` bean twice, if both server and client are in the same application.

I have never used sleuth so, I'm not sure, whether the `spring.sleuth.scheduled.enabled` property check was added intentionally or to avoid conflicts with missing beans. If it is for the later, then maybe we can remove that as well, with the new condition layout?


yidongnan:
We may be able to provide a configuration ``spring.sleuth.grpc.scheduled.enabled`` to open or close the trace.

Issue #132, opened 2018-11-14
Suggestion: Change java package structure


ST-DDT:
Change the package structure to have a common prefix: `net.devh.spring.grpc.` plus create more sub packages for different things.

### 1. Dismatching package infixes

Currently the packages always use different infixes which makes it hard to navigate packages.

#### Examples

* **`net.devh.springboot.autoconfigure.grpc.client`**
* **`net.devh.springboot.autoconfigure.grpc.server`**
* `net.devh.examples.grpc.cloud`
* `net.devh.examples.grpc.local`
* `net.devh.tests.grpc`

Also the package names are quite verbose.

#### Suggestion

Use the same package prefix everywhere and shorten it:

* **`net.devh.spring.grpc.client`**
* **`net.devh.spring.grpc.server`**
* `net.devh.spring.grpc.examples.cloud`
* `net.devh.spring.grpc.examples.local`
* `net.devh.spring.grpc.tests`

### 2. Create more sub-packages

Currently the packages contain many different types of classes, which makes it harder to find specific ones. ~7 classes per package (59 / 8)

#### Examples

* `net.devh.springboot.autoconfigure.grpc.client`
  * `GrpcChannelFactory`s
  * `Interceptor`s
  * `Properties`s
  * `Client` annotation and post processor
  * `AutoConfiguration`

#### Suggestion

Create a package for each of these different categories. 3.5 classes per package (57 / 16)

#### Now

![now_slim](https://user-images.githubusercontent.com/1579362/48518616-7ca27880-e86a-11e8-8e91-3d6926a7fddd.png)

#### Restructured

![restructured_slim](https://user-images.githubusercontent.com/1579362/48518635-888e3a80-e86a-11e8-897b-9cb5a8c50e3a.png)

Note: This way of (re-)structuring is just a suggestion and nothing final.

----------------------

### Pros

* Shorter package names
* More predictable package names
* Classes are easier to find/classes aren't mixed with others
* Future additions can easily follow this strategy
* Side effect: You can easily split the autoconfiguration part to a separate module and provide a lib that works without Spring-Boot.

### Cons

* Breaking change (changed java-package names)
* Deeper package structure (if empty packages are omited)

------------------------------

What do you thing about this suggestion?

yidongnan:
I think the package name is more suitable using ``net.devh.boot.*``, what do you think?
Then I agree with what you said to create more sub-packages.

* net.devh.boot.grpc.client
[...]

PR #135, opened 2018-11-17
Change package structure

ST-DDT:
Implements #132


ST-DDT:
I hope that I fixed this PR properly now.

PR #136, opened 2018-11-19
More secure client auth + share the authorization header constants

ST-DDT:
~~This PR **shouldn't be merged** yet to avoid unnecessary breaking changes. 
[Grpc 1.17](https://github.com/grpc/grpc-java/milestone/30) (which will be released in December) changes the call credentials behavior and thus breaks this code. (`CallCredentials2` will be renamed to/merged with `CallCredentials`)~~
Grpc 1.17 didn't bring the breaking changes and it is uncertain when the break will be. Also It won't be a hard break, since there will be a migration time in the other direction as well. So I guess we can merge this in for now.

### Added features
* Easier to add non-basic-auth features
* Easier to prevent passwords/tokens from being stolen
* Use grpc library feature (`CallCredentials`)

I created the PR this early to allow other developers to comment on this and propose changes. 

### Alternatives considered

* I'm not that happy with the names, so if you have a better suggestion go ahead.
* Instead of the `CallCredentialsAttachingClientInterceptor` it would also be possible to use something like a `StubTransformer` that configures the annotated stubs with the CallCredentials instead of adding it somewhere in the interceptor chain.

### See also

* https://github.com/grpc/grpc-java/issues/1914
* https://github.com/grpc/grpc-java/issues/4901

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientSecurityAutoConfiguration.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientSecurityAutoConfiguration.java
@@ -15,33 +15,52 @@
  * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-package net.devh.boot.grpc.examples.security.client;
+package net.devh.boot.grpc.client.autoconfigure;
 
-import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnSingleCandidate;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 import io.grpc.CallCredentials;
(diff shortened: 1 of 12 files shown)


yidongnan:
I think use something like a ``StubTransformer`` that configures the annotated stubs with the ``CallCredentials`` would be better, but there is also a case where if you use ``@GrpcClient`` to create a ``Channel``, but manually go through ``channel.newBlockingStub`` to generate a stub, then ``StubTransformer`` will can not work

ST-DDT:
Updated and added preview for StubTransformers.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientSecurityAutoConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+ * </p>
+ *
+ * <pre>
+ * <code>@ImportAutoConfiguration(exclude = GrpcClientSecurityAutoConfiguration.class)</code>
+ * </pre>
+ *
+ *
+ * @author Daniel Theuke (daniel.theuke@heuboe.de)
+ */
+@ImportAutoConfiguration(exclude = GrpcClientSecurityAutoConfiguration.class)
```
ST-DDT: TODO: Find out why I added that here. Did I forget to remove after using it as an example?

ST-DDT:
Ready for review.

Issue #106, opened 2018-10-22
Add automated metric support


ST-DDT:
Spring has a module called actuator which allows metric collection.
If we would add support for that, we could check how many messages/requests/responses have been sent, how long it took, and check whether errors have occurred. Server owners could use these metrics and export them to prometheus and visualize them with Grafana or other tools.

![grafana](https://user-images.githubusercontent.com/1579362/47302329-8812d300-d621-11e8-863e-c012197d51da.png)

Are you interested in adding that to this project?
(I might be able to contribute the necessary code for that as well)

yidongnan:
Look forward to your PR

PR #137, opened 2018-11-20
Add support for automatic metric collection

ST-DDT:
Adds #106 

Automatically collect actuator/micrometer metrics for clients and server.


grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/metric/AbstractMetricCollectingInterceptor.java
```diff
@@ -0,0 +1,223 @@
...
+            this.responseCounter = responseCounter;
+            this.timerFunction = timerFunction;
+        }
+
+        /**
+         * Gets the request {@link Counter} for the method related to this instance.
+         *
+         * @return The request counter.
+         */
+        public final Counter getRequestCounter() {
```
yidongnan: You can add the ``@Getter`` annotation to automatically generate the get method.

ST-DDT:
Fixed

yidongnan:
This feature is very useful, thank you for your contribution.

Issue #139, opened 2018-11-21
TLS not working with openJDK8


uweschaefer:
Not sure, if this is the right forum, but i failed to use TLS.

When connecting, i get 

```
Caused by: javax.net.ssl.SSLHandshakeException: General OpenSslEngine problem
	at io.netty.handler.ssl.ReferenceCountedOpenSslContext$AbstractCertificateVerifier.verify(ReferenceCountedOpenSslContext.java:629) ~[netty-handler-4.1.29.Final.jar:4.1.29.Final]
	at io.netty.internal.tcnative.SSL.readFromSSL(Native Method) ~[netty-tcnative-boringssl-static-2.0.20.Final.jar:2.0.20.Final]
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.readPlaintextData(ReferenceCountedOpenSslEngine.java:488) ~[netty-handler-4.1.29.Final.jar:4.1.29.Final]
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.unwrap(ReferenceCountedOpenSslEngine.java:1036) ~[netty-handler-4.1.29.Final.jar:4.1.29.Final]
	at io.netty.handler.ssl.ReferenceCountedOpenSslEngine.unwrap(ReferenceCountedOpenSslEngine.java:1144) ~[netty-handler-4.1.29.Final.jar:4.1.29.Final]
	at io.netty.handler.ssl.SslHandler$SslEngineType$1.unwrap(SslHandler.java:211) ~[netty-handler-4.1.29.Final.jar:4.1.29.Final]
	at io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1297) ~[netty-handler-4.1.29.Final.jar:4.1.29.Final]
	at io.netty.handler.ssl.SslHandler.decodeJdkCompatible(SslHandler.java:1199) ~[netty-handler-4.1.29.Final.jar:4.1.29.Final]
	at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:1243) ~[netty-handler-4.1.29.Final.jar:4.1.29.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:489) ~[netty-codec-4.1.29.Final.jar:4.1.29.Final]
[...]

PR #140, opened 2018-11-22
suggestion: switch to shaded netty dependency, see #139

uweschaefer:
https://github.com/yidongnan/grpc-spring-boot-starter/issues/139

feel free to just drop this if you don't think, it is a good idea.


ST-DDT:
I dont think we should switch to shaded entirely, but maybe we can add the neccessary auto config classes. Then the devs only have to switch their depencies and have still full support from this lib.

uweschaefer:
not sure how to do that, as some of the classes are used in bean provider methods themselves.

yidongnan:
I also think it would be better to do this through automatic configuration. By using the ``@ConditionalOnClass`` annotation to inject the implementation class based on the dependency, but there may be some duplicate code, if there is a way to solve the code duplication here will be better.

ST-DDT:
Do you want to give it a try?

uweschaefer:
i would not know how without significant duplication. probably, some else should open a different PR for that.

ST-DDT:
Thanks for your contribution anyway. I don't know whether I can avoid the duplication myself either.
I will probably try to do that in a few days unless @yidongnan wants to go ahead.

Issue #144, opened 2018-11-27
Publicly host javadocs


ST-DDT:
What do you think about publicly hosting the javadocs?

IIRC there should be some hosters that provide that service for free for open source projects or you can use your own page for that.

yidongnan:
This is a good suggestion, we can first find out which service providers provide hosting services, and whether it can automatically update javadoc through CI.

ST-DDT:
How about this page:
https://www.javadoc.io/

It automatically host javadoc pages for all artifacts on maven central

[![Client-Javadocs](https://www.javadoc.io/badge/net.devh/grpc-client-spring-boot-autoconfigure.svg?label=Client-Javadoc)](https://www.javadoc.io/doc/net.devh/grpc-client-spring-boot-autoconfigure)
[![Server-Javadocs](https://www.javadoc.io/badge/net.devh/grpc-server-spring-boot-autoconfigure.svg?label=Server-Javadoc)](https://www.javadoc.io/doc/net.devh/grpc-server-spring-boot-autoconfigure)

````md
[![Client-Javadocs](https://www.javadoc.io/badge/net.devh/grpc-client-spring-boot-autoconfigure.svg?label=Client-Javadoc)](https://www.javadoc.io/doc/net.devh/grpc-client-spring-boot-autoconfigure)
[![Server-Javadocs](https://www.javadoc.io/badge/net.devh/grpc-server-spring-boot-autoconfigure.svg?label=Server-Javadoc)](https://www.javadoc.io/doc/net.devh/grpc-server-spring-boot-autoconfigure)
````

yidongnan:
This looks nice

ST-DDT:
Should I create a PR for that or do you want to commit it directly?
Should we include a link to common preemptively?

yidongnan:
I submit directly and add common link

ST-DDT:
Looks good. If we add more badges, then we might have to split it to multiple lines.

PR #145, opened 2018-11-27
Automatically create actuator InfoContributor

ST-DDT:
Automatically create and actuator InfoContributor, that exposes the grpc server port and the running services.

actuator-info-example

## Features

* Exposes the grpc server port
* Exposes a sorted map of services with their sorted methods

## TBD

* Should we add a separate property to enable the InfoContributor?
* Should we add a separate property to toggle the service listing?
* What do you thing about adding package-info.java files? (Related #144)
  * I will add the other missing files in a different PR.

package-info-example


yidongnan:
LGTM, I prefer to provide a configuration that allows the user to turn it off. We can set it to the on state by default.
Package-info.java looks very useful for documentation.We can put other missing files

PR #148, opened 2018-11-28
Added tests for streaming calls

ST-DDT:
## Features

* Added `AnonymousAuthenticationReader` to allow some calls to skip authentication, when using AuthorizationCheckingServerInterceptor.
* Added some convenience methods to `AccessPredicate`
* Added tests for streaming calls
* Fixed missing `AccessDeniedException` -> `AuthenticationException` transformations

## TBD

* I noticed that the `permitAll` access setting does not allow unauthenticated users to access the methods, if anonymous authentication is not enabled. Should I add/change it to an explicit `permitAll`?

### DIFF
--- a/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/security/authentication/AnonymousAuthenticationReader.java
+++ b/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/security/authentication/AnonymousAuthenticationReader.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2016-2018 Michael Zhang <yidongnan@gmail.com>
+ *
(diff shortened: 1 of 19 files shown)


yidongnan:
I think you can add/change it to an explicit ``permitAll`` or explain it clearly in the readme, otherwise some issues may appear.

ST-DDT:
I checked the current spring-provided structure and AFAICT it is not possible to add an explicit "permitAll". Thus, I changed the logic a bit to use `null` as public access. However, I'm not sure whether `null` is a good value to disable security. As an alternative I thought about adding a `setPublic(MethodDescriptor)` method to `ManualGrpcSecurityMetadataSource`. What do you think?

ST-DDT:
I guess I'll use a constant for now.

yidongnan:
LGTM

PR #152, opened 2018-12-05
Support java 11 compile

### LINKED ISSUE #146: Also build with Java 11 to ensure compatiblity
Java 8 will be EOL very soon, so we should configure travis to also run the tests with java 11 (OpenJDK because of LTS support?).

Since we are a library it is probably fine to continue to compile to Java 8, but we should ensure that we are Java 11 compatible.

I already tried running the build process with Java 11, but I got some strange errors that are probably gradle related. I also tested using the latest release with an Java 11 Runtime, that worked perfectly fine for me.

yidongnan:
* Gradle 5.0 adjusted optional dependency in gradle/gradle#5434, So part of the dependency we need to add manually.
*  javax.annotation.Generated not found in java 9 above  grpc/grpc-java#3633

fixed #146

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,34 @@
 language: java
-dist: precise
-jdk:
-- oraclejdk8
+
+matrix:
+  include:
+    - jdk: openjdk8
+      env: SKIP_RELEASE=false
+    - jdk: openjdk11
+      env: SKIP_RELEASE=true
+
+branches:
+  #Don't build tags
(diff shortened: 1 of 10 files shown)


yidongnan:
PS: Here is a bug in intellij idea https://youtrack.jetbrains.com/issue/IDEA-201974, It will be fix in the next version``2018.3.1``, but it does not affect the compile and run of the command line.

PR #154, opened 2018-12-06
Update to grpc 1.17.0

ST-DDT:
Updates grpc to 1.17.1.

Requires further testing to verify that it does not break compatibility with pre 1.17.1 versions.

grpc 1.17.1 is compatible with grpc-spring-boot-starter pre 1.17.1.
grpc-spring-boot-starter with 1.17.1 is very likely to be compatible with grpc pre 1.17.1.

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java
@@ -37,7 +37,6 @@ import io.grpc.CompressorRegistry;
 import io.grpc.DecompressorRegistry;
 import io.grpc.LoadBalancer;
 import io.grpc.NameResolver;
-import io.grpc.util.RoundRobinLoadBalancerFactory;
 import net.devh.boot.grpc.client.channelfactory.GrpcChannelConfigurer;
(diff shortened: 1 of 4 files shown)


grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java
```diff
@@ -88,8 +87,10 @@ public AnnotationGlobalClientInterceptorConfigurer annotationGlobalClientInterce
     @ConditionalOnMissingBean
     @Lazy // Not needed for InProcessChannelFactories
     @Bean
+    @SuppressWarnings("deprecation") // Required to stay compatible with pre 1.17.0 versions
     public LoadBalancer.Factory grpcLoadBalancerFactory() {
-        return RoundRobinLoadBalancerFactory.getInstance();
+        return io.grpc.util.RoundRobinLoadBalancerFactory.getInstance();
+        // return LoadBalancerRegistry.getDefaultRegistry().getProvider("round_robin");
```
ST-DDT: This will break in a future version of grpc, but I cannot update it because that would break backwards compatibility.

yidongnan:
I see the grpc release page saying don't use this version.
> Do not use this release. Its handling of gRPCLB causes problems for the ecosystem

ST-DDT:
Updated to 1.17.1

PR #157, opened 2018-12-07
Dynamic name resolver factory

### LINKED ISSUE #153: Allow dynamic selection of NameResolverFactory
Currently it is only possible to declare a single `NameResolver.Factory` regardless how many different targets you want to communicate with. However, if you have multiple targets then you might also have different ways to communicate with them. For example: The internal services might be available via `DiscoveryClient` but the external services should be resolved via DNS.

It would be nice, if the developer could define the target address dynamically and the service implementation will decide how to resolve that name.

## Example-Config
````properties
grpc.client.serviceA.address=dns://foo.bar.example:4000  # grpc native
grpc.client.serviceB.address=discovery://serviceB.spring.name
grpc.client.serviceC.address=static://127.0.0.1:9090,127.0.0.1:9091 # currently referred to as AddressChannelNameResolver
````

What do you think of this suggestion?

PS: I already have a rough idea how to implement this, but it might take a few days before I can create a PR for that.

### PR DESCRIPTION
**Pull request by ST-DDT**
Fixes #153 

This PR contains the following fixes/changes:

* Added `grpc.client.<name>.address`
  * `dns:/foo.bar.example:4000/`
  * `static://localhost:4000`
  * `static://10.0.0.5:9090,10.0.0.5:9090`
  * `discovery://foobar` <-- Not yet tested
* Removed `grpc.client.<name>.host` and `grpc.client.<name>.port`
* Reduced `DiscoveryClientResolverFactory` to `NameResolver.Factory`, because it cannot be loaded by java's service registry and thus the priority/availability flags won't apply.
* Fixed bug in `DiscoveryClientResolverFactory` that it didn't check the scheme in the uri.
* Moved `grpc.client`'s address resolution from the actual name resolvers to a Spring/config abstraction
* Reimplemented `StaticNameResolverFactory` and load it via java's service registry along with the grpc/other library provided implementations
* Change client to try to connect to `localhost` instead of `127.0.0.1` by default
* Change server to also listen to IPv6 by default (was IPv4 only)

### CODE REVIEW

File: grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/config/GrpcChannelProperties.java [hunk 1]
```diff
@@ -33,24 +32,49 @@
...
+    public static final String DEFAULT_HOST = "localhost";
+    public static final int DEFAULT_PORT = 9090;
+    public static final URI DEFAULT_URI = URI.create("static://" + DEFAULT_HOST + ':' + DEFAULT_PORT);
 
     public static final GrpcChannelProperties DEFAULT = new GrpcChannelProperties();
 
     /**
-     * A list of hosts to connect to. These entries should be kept in tandem with the port entries. Defaults to
-     * {@link #DEFAULT_HOST}.
+     * The target uri in the format: 'schema:/host/' or 'schema://authority'. Defaults to 'static://localhost:9090'.
```
ST-DDT: I have no clue, what the difference between `'schema:/host/'` and `'schema://authority'` is.

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/config/GrpcServerProperties.java
```diff
@@ -35,10 +35,15 @@
 @ConfigurationProperties("grpc.server")
 public class GrpcServerProperties {
 
+    public static String ANY_IP_ADDRESS = "*";
```
ST-DDT: Is this a good way to declare this or is there a better constant out there?

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/config/GrpcChannelProperties.java
ST-DDT: I made a mistake here

ST-DDT: Fixed

yidongnan:
The config file in the example project needs to be modified.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/config/GrpcChannelProperties.java
```diff
@@ -33,24 +32,68 @@
...
+     * <p>
+     * <b>Examples</b>
+     * </p>
+     *
+     * <ul>
+     * <li>{@code static://localhost:9090} (refers to exactly one IPv4 or IPv6 address, dependent on the jre
+     * configuration, it does not check whether there is actually someone listening on that network interface)</li>
+     * <li>{@code static://10.0.0.10}</li>
+     * <li>{@code static://10.0.0.10,10.11.12.11}</li>
+     * <li>{@code static://10.0.0.10:9090,10.0.0.11:80,10.0.0.12:1234,[::1]:8080}</li>
```
yidongnan: Do we need to separate ipv4 from ipv6? What do you think?
https://github.com/grpc/grpc/blob/master/doc/naming.md

ST-DDT: I haven't seen that page at all yet. Our system probably also supports those ways to contact the server. But I couldn't find any implementation classes related to those variants so I'm not sure. 

Anyway, our advantage over their impls is that we allow mixed connections to both ipv4 and ipv6 in one address/without modifying the schema. Maybe we can push the mixed version upstream or rename the schema to `ip` or `ipvx`. Another alternative would be using multiple address blocks, but the ipv4/ipv6 combination is probably the only case where that makes any sense.

yidongnan:
``cloud-grpc-client`` project needs to add a configuration ``address: discovery:///cloud-grpc-server``.
The [spring.factories](https://github.com/yidongnan/grpc-spring-boot-starter/blob/9d64bdbb39008c53c0d44a12471d9020729a0abf/grpc-client-spring-boot-autoconfigure/src/main/resources/META-INF/spring.factories#L5) need to add ``net.devh.boot.grpc.client.autoconfigure.GrpcDiscoveryClientAutoConfiguration``

ST-DDT:
Oh yeah, you are right. I totally missed that.

As for `address: discovery:///cloud-grpc-server` we could change the logic a bit so it will be derived automatically.

For example we could remove the default value from the properties and use a configurable `Function<clientName, addressUri>` in the `ConfigMappedNameResolverFactory`  if no address is set.

````java
private final Function<String, URI> defaultUriMapper;

@Nullable
@Override
public NameResolver newNameResolver(final URI targetUri, final Attributes params) {
    final String clientName = targetUri.toString();
    final GrpcChannelProperties clientConfig = this.config.getChannel(clientName);
    URI remappedUri = clientConfig.getAddress();
    if (remappedUri == null) {
        remappedUri = defaultUriMapper.apply(clientName);
    }
    log.debug("Remapping target URI for {} to {} via {}",
            clientName, remappedUri, this.delegate);
    final Attributes extendedParas = params.toBuilder()
            .set(NameResolverConstants.PARAMS_CLIENT_NAME, clientName)
            .set(NameResolverConstants.PARAMS_CLIENT_CONFIG, clientConfig)
            .build();
    return this.delegate.newNameResolver(remappedUri, extendedParas);
}
````

yidongnan:
This looks great, so it don't need to configure each client. Will you add this feature to this PR?

ST-DDT:
I will add it to this PR.

PR #166, opened 2018-12-28
Add links to other public javadoc pages

ST-DDT:
Adds links to other public javadoc pages.

**This has the following benefits**

* Replaces fqcn with links to the official docs, which are usually more useful then just the name.
* Allows creating proper hierarchies for custom implementations (?)
* Derive some javadocs from the external parents (?)

**Limitations**

* Does not work with java 11
  * This might be caused by a bad configuration of grpc-java's javadocs (https://github.com/grpc/grpc.github.io/issues/801)

**TBD/TBC**

* Do we want the external links at all?
* Currently the javadoc links for spring reference a specific version of spring which is not bound to the dependency version. Is it possible to somehow reference the variables instead?
* Need to check which version of the spring-framework is actually used in the referenced spring-boot lib.


yidongnan:
> Do we want the external links at all?

I think this is ok.

> Currently the javadoc links for spring reference a specific version of spring which is not bound to the dependency version. Is it possible to somehow reference the variables instead?
Need to check which version of the spring-framework is actually used in the referenced spring-boot lib.

We can read the version of the dependency from the dependency manager

ST-DDT:
>> Currently the javadoc links for spring reference a specific version of spring which is not bound to the dependency version. Is it possible to somehow reference the variables instead?
    Need to check which version of the spring-framework is actually used in the referenced spring-boot lib.
> 
> We can read the version of the dependency from the dependency manager

Do you know how to do this? Does it work like [this](https://gist.github.com/dudadornelles/be0bea9f98d3aa17404c)?

Issue #177, opened 2019-01-24
Ability to configure server keepalive?


internetstaff:
I see support for configuring client keepalives, but not server keepalives.  Am I just missing it?
Thanks!

internetstaff:
Taking a second pass through the code, I found GrpcServerConfigurer.  Very nice! I may submit a PR with at least a documentation update if not support for configuring server keepalives in the config.

ST-DDT:
Fixed via #185

PR #178, opened 2019-01-25
Adds GrpcServerConfigurer example to README.

internetstaff:
Hopefully this stops questions like mine #177 and provides a useful example of configuring server-side keepalives for those of us struggling with AWS network load balancers. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,25 @@ config) to your dependencies and then configure it as needed.
   }
   ````
 
+#### Customizing a Server
+
+This library also supports custom changes to the ServerBuilder during creation by creating GrpcServerConfigurer beans.
+
+````java
+  @Bean
+  public GrpcServerConfigurer keepAliveServerConfigurer() {
+    return serverBuilder -> {
+      if (serverBuilder instanceof NettyServerBuilder) {
+        ((NettyServerBuilder)serverBuilder)
+            .keepAliveTime(30, TimeUnit.SECONDS)
+            .keepAliveTimeout(5, TimeUnit.SECONDS)
+            .permitKeepAliveWithoutCalls(true);
+      }
+    };
+  }
+````
+
+
 ### gRPC client
 
 To add a dependency using Maven, use the following:
--- a/README.md
+++ b/README.md
@@ -110,6 +110,24 @@ grpc.server.port=9090
 grpc.server.address=0.0.0.0
 ````
 
+#### Customizing a Server
+
+This library also supports custom changes to the `ServerBuilder` during creation by creating `GrpcServerConfigurer` beans.
+
+````java
+@Bean
+public GrpcServerConfigurer keepAliveServerConfigurer() {
+  return serverBuilder -> {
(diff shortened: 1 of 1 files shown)


yidongnan:
Thank you for your commit. I think we can provide automatic configuration in the server, just like the client side.
Are you interested in adding that to this project?

internetstaff:
@yidongnan Yes, I'm interested in tackling that. I'm not sure when I can get to it, but it's on my todo list. :)

@ST-DDT done, PR updated.

README.md
```diff
@@ -223,6 +223,25 @@ config) to your dependencies and then configure it as needed.
   }
   ````
 
+#### Customizing a Server
+
+This library also supports custom changes to the ServerBuilder during creation by creating GrpcServerConfigurer beans.
```
ST-DDT: Use inline `code` blocks to highlight actual classes and interfaces.

README.md
```diff
@@ -223,6 +223,25 @@ config) to your dependencies and then configure it as needed.
...
+  public GrpcServerConfigurer keepAliveServerConfigurer() {
+    return serverBuilder -> {
+      if (serverBuilder instanceof NettyServerBuilder) {
+        ((NettyServerBuilder)serverBuilder)
+            .keepAliveTime(30, TimeUnit.SECONDS)
+            .keepAliveTimeout(5, TimeUnit.SECONDS)
+            .permitKeepAliveWithoutCalls(true);
+      }
+    };
+  }
```
ST-DDT: Remove the outermost indentation, as it does not help readability and developers have to run it throught the auto formatter anyway.

README.md
```diff
@@ -223,6 +223,25 @@ config) to your dependencies and then configure it as needed.
   }
   ````
 
+#### Customizing a Server
```
ST-DDT: I would move this section above the server security section (after the properties section), because it is much smaller.

README.md
```diff
@@ -223,6 +223,25 @@ config) to your dependencies and then configure it as needed.
...
+
+This library also supports custom changes to the ServerBuilder during creation by creating GrpcServerConfigurer beans.
+
+````java
+  @Bean
+  public GrpcServerConfigurer keepAliveServerConfigurer() {
+    return serverBuilder -> {
+      if (serverBuilder instanceof NettyServerBuilder) {
+        ((NettyServerBuilder)serverBuilder)
```
ST-DDT: Please add a space between the cast and the variable.

ST-DDT:
Maybe we could add an example for the client side as well? `GrpcChannelConfigurer` and `StubTransformer`

@yidongnan Since the documentation is now getting more and more extensive, should we start to divide it into several documents or is the current approach with everything in one better? (Not as part of this PR)

yidongnan:
> Since the documentation is now getting more and more extensive, should we start to divide it into several documents or is the current approach with everything in one better? (Not as part of this PR)

Divide it into several documents  and provide a document index, it is more intuitive.

internetstaff:
I think I've addressed everything.

PR #182, opened 2019-01-31
Upgrade to grpc 1.18 + use shaded netty by default

ST-DDT:
* Upgrade to grpc 1.18
* Use shaded netty
* Update documentation
* Remove some unneeded code
* Update to Spring-Boot 2.1 (backwards compatible)
* Update to Spring-Cloud Greenwich (backwards compatible)


yidongnan:
I am so sorry that I have come to reply now. During the Chinese New Year, I went to a small village where the mobile network was very bad and there was no way to respond in time. I will handle the to-do list as soon as possible, thanks

ST-DDT:
No problem. Everyone needs a break from time to time.

Issue #183, opened 2019-02-01
Is there a global config for all servers?


andcuevas:
I would like to know if there is an alternative way to configure all the servers on the client side, instead of using grpc.client.$server_name.some_property
I'm working in a project where the list of server names might be big and I would like to configure all the servers with one property or if this is not possible maybe do it in the code.

ST-DDT:
Currently there is no way to do this via property/config. However, I plan to add this feature. Do you have a suggestion how to add/implement such a feature?

There is a way to do it via code though: [`GrpcChannelConfigurer`](https://static.javadoc.io/net.devh/grpc-client-spring-boot-autoconfigure/2.2.1.RELEASE/net/devh/boot/grpc/client/channelfactory/GrpcChannelConfigurer.html)

andcuevas:
You have default properties in GrpcChannelProperties.DEFAULT.
If that instance could first look the presence of a global property, if the global property is not there use the default that is hard coded.

ST-DDT:
Work in progress.

ST-DDT:
See #198 for the implementation

PR #185, opened 2019-02-03
Add support for server keepAlives + permit(Client)KeepAlives. Adds #177

ST-DDT:
Adds #177
Related to #183 (but not yet fixed)

* Add support for server keepAlives
* Add support for permit(Client)KeepAlives. 
* Add support for duration units (ex. use 1h instead of 3600 (seconds))

`DataSize` is also planned, but it's only available in Spring 5.1 which we will update to in another PR.


ST-DDT:
@internetstaff I hope this is what you need.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
```diff
@@ -168,9 +168,9 @@ protected void configure(final T builder, final String name) {
     protected void configureKeepAlive(final T builder, final String name) {
         final GrpcChannelProperties properties = getPropertiesFor(name);
         if (properties.isEnableKeepAlive()) {
-            builder.keepAliveWithoutCalls(properties.isKeepAliveWithoutCalls())
-                    .keepAliveTime(properties.getKeepAliveTime(), TimeUnit.SECONDS)
-                    .keepAliveTimeout(properties.getKeepAliveTimeout(), TimeUnit.SECONDS);
+            builder.keepAliveTime(properties.getKeepAliveTime().toNanos(), TimeUnit.NANOSECONDS)
```
internetstaff: nit: seems a hair cleaner to leave it in seconds, but that's subjective.

ST-DDT: Grpc uses nanos internally

PR #187, opened 2019-02-11
Use new DataSize feature to simplify size limit setup

ST-DDT:
This PR changes the configuration to use Spring-Boot 2.1's `DataSize` feature which simplifies the setup of the size limit by using data size unit suffixes.

## Example

**Current**

````properties
grpc.server.maxInboundMessageSize=5242880
````

**New**

````properties
grpc.server.maxInboundMessageSize=5MB
````

(The same works also for the client)

## Pros

Simplified configuration

## Cons

Spring Boot 2.0 cannot be used anymore, because this change requires features from Spring-Boot 2.1 or later. It might be possible to still use 2.0 by manually copying the two required classes from 2.1 (not tested).

## Tasks

* [x] Use `DataSize` feature
* [x] Add tests to ensure config backwards compatibility
* [x] Discus whether this change is worth its price

### DIFF
--- a/README-zh.md
+++ b/README-zh.md
@@ -34,9 +34,11 @@ Java技术交流群：294712648 <a target="_blank" href="http://shang.qq.com/wpa
 
 ## 版本
 
-2.x.x.RELEASE 支持 Spring Boot 2 & Spring Cloud Finchley, Greenwich。
+2.x.x.RELEASE 支持 Spring Boot 2.1+ & Spring Cloud Greenwich。
 
 最新的版本：``2.4.0.RELEASE``
+
+(使用 `2.4.0.RELEASE` 版本可以支持 Spring Boot 2.0.X & Spring Cloud Finchley).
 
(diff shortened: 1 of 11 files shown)


yidongnan:
LGTM, this change makes the configuration clearer, but it need to force the user to upgrade the spring-boot to 2.1.x, I think it can be merged later.

ST-DDT:
[Spring 5.0 is EOL (superseded by 5.1)](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Versions#supported-versions) and I assume that the same is true for Spring Boot 2.0, but I couldn't find any info related to that topic.

Should we continue waiting or do we drop support for SB2.0? (It does not matter to me)

yidongnan:
drop support for SB2.0 😄

ST-DDT:
@yidongnan I updated the PR. Now it's ready for review.

PR #195, opened 2019-02-21
Move Trace configuration to its own file

### LINKED ISSUE #194: Problem with client autoconfiguration
Not sure if I'm missing something bout for some reason the grpc tracing interceptor is not being created and I can't see why:

GrpcClientAutoConfiguration.TraceClientAutoConfiguration:
      Did not match:
         - @ConditionalOnBean (types: brave.grpc.GrpcTracing; SearchStrategy: all) did not find any beans of type brave.grpc.GrpcTracing (OnBeanCondition)
      Matched:
         - @ConditionalOnProperty (spring.sleuth.scheduled.enabled) matched (OnPropertyCondition)

However, I can see the bean created:

GrpcCommonTraceAutoConfiguration matched:
      - @ConditionalOnClass found required classes 'brave.Tracing', 'brave.grpc.GrpcTracing' (OnClassCondition)
      - @ConditionalOnProperty (spring.sleuth.scheduled.enabled) matched (OnPropertyCondition)

The only thing I can think of is that the bean is created after the TraceClientAutoConfiguration is being evaluated but I don't see how is this possible.

Am I missing something here?

The Server side is being created successfully.

Any help is appreciated,

Thanks

ST-DDT:
**Fixes #194**

The ordering annotations on nested auto configuration classes will be ignored, thus the might be executed in the wrong order. This PR solves the issue by moving all auto configuration classes to their own classes.

Thanks @ramses-gomez for reporting this issue.


ramses-gomez:
Thanks a lot for fixing this issue so fast, I was doing what you suggested, it is good to know that order is ignored in nested autoconfiguration classes. Thanks a lot again

PR #198, opened 2019-02-27
Add support for global client properties

ST-DDT:
Adds #183

This PRs adds the `grpc.client.GLOBAL` client configuration that will be used as fallback, if the properties are not configured per client.

Order of precedence: Per Client > `GLOBAL` > defaults

I also attempted to add some metadata for the client side. Unfortunately it still does not show the per client details, but you can use the `GLOBAL` properties as a reference. In eclipse this works up to a certain degree. 

eclipse-preview

I'll test it with IntelliJ IDEA in a few days, but I'm not familiar with it, so can someone else test it with IntelliJ as well? 

Thanks to @andcuevas for suggesting this feature.

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/grpc-client-spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -32,7 +32,7 @@
 			"name": "grpc.client.GLOBAL.address",
 			"type": "java.net.URI",
 			"sourceType": "net.devh.boot.grpc.client.config.GrpcChannelProperties",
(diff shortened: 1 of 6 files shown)


grpc-client-spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -8,7 +8,7 @@
 		},
 		{
 			"name": "grpc.client",
-			"type": "java.util.Map<java.util.String,net.devh.boot.grpc.client.config.GrpcChannelProperties>",
+			"type": "java.util.Map<java.lang.String,net.devh.boot.grpc.client.config.GrpcChannelProperties>",
```
ST-DDT: Good finding

ST-DDT:
@yidongnan Sorry, I accidentally dismissed your review by pushing a small typo fix.

PR #204, opened 2019-03-21
Update example dependencies to fix compilation error

ST-DDT:
The previously used `BUILD-SNAPSHOT` version cause compilation errors, this PR updates them to a milestone release and thus fixes the build errors.


yidongnan:
The last compilation error was because the `repo.spring.io`  is unavailable.
https://twitter.com/springops/status/1105504213390450689
I don't think there is any need to change it now, unless the RELEASE version appears.

ST-DDT:
But why did I get a compilation error if a repo is down? I could understand it, if it were a dependecy missing error.

(The Milestone releases are less likely to change in the future.)

PR #208, opened 2019-04-03
Make GrpcServerSecurityAutoConfiguration a proper auto configuration

### LINKED ISSUE #205: Cannot disable autoconfiguration for spring security
It doesn't seem possible to opt out from the auto configuration in `GrpcSecurityAutoConfiguration` (conditional on `AuthenticationManager` bean).

I tried to add `GrpcSecurityAutoConfiguration` to the list of excludes for auto configuration but spring boot complaints
> The following classes could not be excluded because they are not auto-configuration classes

(I tried both using `spring.autoconfigure.exclude` property and directly in `@SpringBootApplication(exclude ... )` 

The application in question is using spring security but it won't be using it for the gRPC parts.

ST-DDT:
Fixes #205 

Instead of importing the configuration, it is referenced via `spring.factories` this enables the users of the library to explicitly exclude the auto configuration from applying.

### DIFF
--- a/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcServerSecurityAutoConfiguration.java
+++ b/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcServerSecurityAutoConfiguration.java
@@ -15,15 +15,18 @@
  * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-package net.devh.boot.grpc.server.security;
+package net.devh.boot.grpc.server.autoconfigure;
 
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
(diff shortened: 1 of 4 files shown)


grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcServerSecurityAutoConfiguration.java
```diff
@@ -54,16 +57,19 @@
...
+@AutoConfigureAfter(WebSecurityEnablerConfiguration.class)
+public class GrpcServerSecurityAutoConfiguration {
 
     @Bean
     @ConditionalOnMissingBean
     @ConditionalOnBean({AccessDecisionManager.class, GrpcSecurityMetadataSource.class})
     public AuthorizationCheckingServerInterceptor authorizationCheckingServerInterceptor(
             final AccessDecisionManager accessDecisionManager,
             final GrpcSecurityMetadataSource securityMetadataSource) {
+        log.warn("Create AuthorizationCheckingServerInterceptor");
```
yidongnan: I don't understand the meaning of the `warn` log here.

ST-DDT: Looks like I forgot to remove a debug comment

PR #213, opened 2019-04-10
Order interceptors based on Spring's Order annotation

rajki:
Addresses #214 

Both server and client interceptors will now be registered based on the order specified by the `@Order` annotation. Interceptors without the annotation will be registered first. Annotated interceptors will be registered in ascending order of precedence.

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/interceptor/AnnotationGlobalClientInterceptorConfigurer.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/interceptor/AnnotationGlobalClientInterceptorConfigurer.java
@@ -17,15 +17,12 @@
 
 package net.devh.boot.grpc.client.interceptor;
 
+import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


ST-DDT:
Did you close this PR by mistake?

rajki:
@ST-DDT Hi! Initially, I was fetching by `@Order` annotation and sorting them, but then realised that it could be done in a better manner by using `AnnotationAwareOrderComparator`. So, closed this PR until I make the required changes, and do a round of testing with `@Order`, `@Priority` annotations, and by extending `Ordered` interface.

If you feel accommodating `@Order` annotation alone is sufficient, I'll open a new PR with the initial goal. Else will complete the set of changes and then raise a new PR.

ST-DDT:
If we add support for sorting them then we should do it with complete support as you suggested.

Maybe we could use this feature to sort the custom per client/server interceptors specified via annotations as well. Potentially with another annotation field like `sortInterceptors: boolean = false`.

I'm looking forward to your contribution.

rajki:
@ST-DDT raised a new PR here- #215

PR #223, opened 2019-04-26
Adds configurable ciphers and protocols for clients and servers

NickUfer:
This PR adds support for specifying which protocols and ciphers should be allowed for secure connections. This works for both clients and servers

---
### Configuration

Both the protocols and the ciphers must be specified as a list, separated by either colons, spaces or commas. Colons, commas and spaces could be mixed.

**Valid out of the ordinary examples for protocol lists and cipher suites (lists)**:

```properties
grpc.server.security.protocols="TLSv1.1:TLSv1.2"
grpc.server.security.ciphers="ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305,ECDHE-RSA-AES128-SHA"
grpc.client.myClient.security.protocols="TLSv1.1,TLSv1.2 "
grpc.client.myClient.security.ciphers="ECDHE-ECDSA-AES128-GCM-SHA256,ECDHE-ECDSA-CHACHA20-POLY1305 ECDHE-RSA-AES128-SHA"
```

**Example Server Configuration**
```properties
grpc.server.security.protocols="TLSv1.1:TLSv1.2"
grpc.server.security.ciphers="ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-AES128-SHA"
```

**Example Client Configuration**
```properties
grpc.client.myClient.security.protocols="TLSv1.2"
grpc.client.myClient.security.ciphers="ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-ECDSA-CHACHA20-POLY1305"
```

### Behaviour if not specified

If no or an empty cipher suite or protocol list was specified, the system's default protocols or cipher suite will be used. So this is no breaking change.

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/NettyChannelFactory.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/NettyChannelFactory.java
@@ -104,15 +104,11 @@ public class NettyChannelFactory extends AbstractChannelFactory<NettyChannelBuil
             }
 
             if (security.getCiphers() != null && !security.getCiphers().isEmpty()) {
-                // splits the cipher list at colons, commas or spaces
-                String[] extractedCiphers = security.getCiphers().split("[:, ]");
-                sslContextBuilder.ciphers(Arrays.asList(extractedCiphers));
+                sslContextBuilder.ciphers(security.getCiphers());
             }
 
-            if (security.getProtocols() != null && !security.getProtocols().isEmpty()) {
-                // splits the protocol list at colons, commas or spaces
-                String[] extractedProtocols = security.getProtocols().split("[:, ]");
-                sslContextBuilder.protocols(extractedProtocols);
+            if (security.getProtocols() != null && security.getProtocols().length > 0) {
+                sslContextBuilder.protocols(security.getProtocols());
             }
 
             try {
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/NettyChannelFactory.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/NettyChannelFactory.java
@@ -20,7 +20,6 @@ package net.devh.boot.grpc.client.channelfactory;
 import static java.util.Objects.requireNonNull;
 
 import java.io.File;
-import java.util.Arrays;
 import java.util.List;
 
 import javax.net.ssl.SSLException;
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/NettyChannelFactory.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/NettyChannelFactory.java
@@ -103,13 +103,13 @@ public class NettyChannelFactory extends AbstractChannelFactory<NettyChannelBuil
                 sslContextBuilder.trustManager(trustCertCollectionFile);
             }
 
-            if(security.getCiphers() != null && !security.getCiphers().isEmpty()) {
+            if (security.getCiphers() != null && !security.getCiphers().isEmpty()) {
                 // splits the cipher list at colons, commas or spaces
                 String[] extractedCiphers = security.getCiphers().split("[:, ]");
                 sslContextBuilder.ciphers(Arrays.asList(extractedCiphers));
             }
 
-            if(security.getProtocols() != null && !security.getProtocols().isEmpty()) {
+            if (security.getProtocols() != null && !security.getProtocols().isEmpty()) {
                 // splits the protocol list at colons, commas or spaces
                 String[] extractedProtocols = security.getProtocols().split("[:, ]");
                 sslContextBuilder.protocols(extractedProtocols);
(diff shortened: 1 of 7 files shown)


grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/NettyChannelFactory.java
```diff
@@ -102,6 +103,18 @@ protected void configureSecurity(final NettyChannelBuilder builder, final String
                 sslContextBuilder.trustManager(trustCertCollectionFile);
             }
 
+            if (security.getCiphers() != null && !security.getCiphers().isEmpty()) {
+                // splits the cipher list at colons, commas or spaces
+                String[] extractedCiphers = security.getCiphers().split("[:, ]");
```
ST-DDT: Have you considered/tried making the ciphers/protocols a `List<String>` or `String[]` directly?
IMO it should still be possible to configure it in the same way then (except for the other separators). Spring is usually smart in these things.
If you want to also support the other separators you can add an extra setter for single strings.
That way we won't leak the configuration definition into our main code.

NickUfer: Yeah I also thought about that. IMO the problem here is that cipher suites are usually specified with colons as separators but I also wanted to support the commas for the spring style. 
I found out that I could solve that by using a `@Value` annotation like this `@Value("#{'${my.list.of.strings}'.split('[,: ]')}")` but this would not work with dynamic names as in the client properties.
Maybe I should move the splitting to the getters to return either `List<String>` or `String[]`?

NickUfer:
> It would be nice, if we had tests for these properties.
> Could you add some or shall I?

Tests will be added asap

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/NettyChannelFactory.java
ST-DDT: > Yeah I also thought about that. IMO the problem here is that cipher suites are usually specified with colons as separators but I also wanted to support the commas for the spring style.

It might be possible to define two setters; one with just a `String` and one with a `List<String>`.
Maybe Spring is smart enough to figure out which one it should use.

> Maybe I should move the splitting to the getters to return either List<String> or String[]?

I would add the splitting logic into the setter and go for `List<String>` instead of an array.

NickUfer: > It might be possible to define two setters; one with just a String and one with a List<String>.

I've tried that approach with 2 setters, one with `List<String>` and one just with `String` as the parameter, but with `List<String>` or `String[]` spring just adds the whole entry and doesn't split them correctly. So I decided to move the splitting into the normal setter and remove the other ones.

tests/src/test/java/net/devh/boot/grpc/test/CustomCiphersAndProtocolsSetupTest.java
```diff
@@ -0,0 +1,136 @@
...
+    private TestServiceGrpc.TestServiceBlockingStub tlsV12Stub;
+    @GrpcClient("tls13")
+    private TestServiceGrpc.TestServiceBlockingStub tlsV13Stub;
+    @GrpcClient("noSharedCiphers")
+    private TestServiceGrpc.TestServiceBlockingStub tlsNoSharedCiphersStub;
+    @GrpcClient("noSharedProtocols")
+    private TestServiceGrpc.TestServiceBlockingStub tlsNoSharedProtocolsStub;
+
+    /**
+     * Tests behaviour with TLSv1.1 and shared protocols. Test should succeed, as the server supports TLSv1.1.
```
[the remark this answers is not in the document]
NickUfer: Done

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/config/GrpcChannelProperties.java
```diff
@@ -550,6 +552,42 @@ public void setAuthorityOverride(final String authorityOverride) {
...
+        /**
+         * @param ciphers Cipher suite consisting of one or more cipher strings separated by colons, commas or spaces
+         */
+        public void setCiphers(String ciphers) {
+            this.ciphers = Arrays.asList(ciphers.split("[ :,]"));
+        }
+
+        // --------------------------------------------------
+
+        private String[] protocols = null;
```
yidongnan: `ciphers` filed is the `List<String>` type, `protocols` field is the `String[]` type, Is it possible to use the same type here?

ST-DDT: IIRC the netty API is like this.
Should we deviate from it to have a more homogenous code ourselves?

yidongnan: Sorry, I missed the netty api. Keeping existing will be better.

ST-DDT:
Thanks for your contribution.

PR #224, opened 2019-05-01
Add result counter tagged with statuscode

david-barrett:
Adds new counters that counts the results of each request and tags with a status code

New counters
`grpc.server.requests.result`
`grpc.client.requests.result`

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/metric/MetricCollectingClientCall.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/metric/MetricCollectingClientCall.java
@@ -40,6 +40,7 @@ class MetricCollectingClientCall<Q, A> extends SimpleForwardingClientCall<Q, A> 
     private final Counter requestCounter;
     private final Counter responseCounter;
     private final Function<Code, Timer> timerFunction;
+    final Function<Code, Counter> resultFunction;
 
     /**
      * Creates a new delegating ClientCall that will wrap the given client call to collect metrics.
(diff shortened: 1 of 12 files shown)


ST-DDT:
Sorry, for being slow with reviewing your PR.

What is the difference/added benefit between the existing `grpc.client.processing.duration` and the new `grpc.client.requests.result`?

**existing**

````js
{
  "name": "grpc.client.processing.duration",
  "description": "The total time taken for the server to process the requests including network delay as observed by the client",
  "baseUnit": "seconds",
  "measurements": [
    {
      "statistic": "COUNT", // <--- here
      "value": 1
    },
    {
      "statistic": "TOTAL_TIME",
      "value": 0.3202857
    },
    {
      "statistic": "MAX",
      "value": 0.3202857
    }
  ],
  "availableTags": [
    {
      "tag": "method",
      "values": [
        "SayHello"
      ]
    },
    {
      "tag": "service",
      "values": [
        "Simple"
      ]
    },
    {
      "tag": "statusCode",
      "values": [
        "OK"
      ]
    }
  ]
}
````

vs.

**added**

````json
{
  "name": "grpc.client.requests.result",
  "description": "The request results as observed by the client",
  "baseUnit": "messages",
  "measurements": [
    {
      "statistic": "COUNT",
      "value": 1
    }
  ],
  "availableTags": [
    {
      "tag": "method",
      "values": [
        "SayHello"
      ]
    },
    {
      "tag": "service",
      "values": [
        "Simple"
      ]
    },
    {
      "tag": "statusCode",
      "values": [
        "OK"
      ]
    }
  ]
}
````

AFAICT the count is already included in the timer stats and at least prometheus exposes them both for consumption.

david-barrett:
No, you are totally correct.  I wasn't aware a timer could also supply a count of requests, so this PR doesn't add any value.  

Thanks for checking
I'll close

ST-DDT:
Thanks for contributing

Issue #214, opened 2019-04-10
Support for Order annotation while registering Server and Client interceptors


rajki:
Currently, the way to ensure ordering of annotations is to write a configurer and add them in the specific order as we'd like them to execute. Alternatively, we could support annotating interceptor classes with `@Order` along with `@GrpcGlobalServerInterceptor`/`@GrpcGlobalClientInterceptor`.

ST-DDT:
Maybe we could use this feature to sort the custom per client/service interceptors specified via annotations as well. Potentially with another annotation field/attribute like `sortInterceptors: boolean = false`.

E.g.:
````java
@GrpcService(interceptors = LoadEstimatingServerInterceptor, sortInterceptors = true)
````

rajki:
Implemented sorting of global interceptors. Haven't done it for service specific interceptors.

rajki:
@ST-DDT thank you for all the help with this, which happens to be my first open source contribution 😄 should I close this issue?

ST-DDT:
You're welcome.
If you have any other suggestion for this project just go ahead and open new issues / PRs.
I leave this issue open while I consider whether the custom interceptor ordering is useful.

PR #233, opened 2019-05-29
Order interceptors

### LINKED ISSUE #226: How to setup a global error handler interceptor
What is the design purpose of calling service specific interceptor before global interceptor?

My use case is that, I want to setup a global interceptor to handle all exceptions from service implementation and other interceptors. But I got interceptor order issues. First I cannot guarantee the global interceptor order, I guess #215 can fix this issue. But I found another issue with service specific interceptor, it seems called before global interceptor.

```
public @interface GrpcService {

    /**
     * A list of {@link ServerInterceptor} classes that should be applied to only this service. If a bean of the given
     * type exists, it will be used; otherwise a new instance of that class will be created via no-args constructor.
     *
     * <p>
     * <b>Note:</b> These interceptors will be applied after the global interceptors. But the interceptors that were
     * applied last, will be called first.
     * </p>
     *
     * @return A list of ServerInterceptor classes that should be used.
     */
    Class<? extends ServerInterceptor>[] interceptors() default {};

    /**
     * A list of {@link ServerInterceptor} beans that should be applied to only this service.
     *
     * <p>
     * <b>Note:</b> These interceptors will be applied after the global interceptors and the interceptor classes. But
     * the interceptors that were applied last, will be called first.
     * </p>
     *
     * @return A list of ServerInterceptor beans that should be used.
     */
    String[] interceptorNames() default {};

}

```

Is there any way to setup a global interceptor that called as the first interceptor and handle all exceptions from service implementation or other interceptor?

I currently rewrite `AnnotationGrpcServiceDiscoverer` to manually adjust the order of interceptors to move my global error handler interceptor to the end of the list and get first called. It works but looks like a hack way.

ST-DDT:
Assigns priorities to the default interceptors and move the sorting of the interceptors to a central place, to make overwriting the logic easier.

Fixes #226 
Adds the last bits for #214

I couldn't find a way to make it less disrupting. 😢 

**Known limitations:**
* `@Order` on `@Bean` methods in `@Configuration` classes won't be picked up by Spring. I haven't found a solution for that yet: https://stackoverflow.com/questions/56331925/sort-spring-beans-by-method-level-order-annotation

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
@@ -104,7 +104,7 @@ public abstract class AbstractChannelFactory<T extends ManagedChannelBuilder<T>>
         if (sortInterceptors) {
             this.globalClientInterceptorRegistry.sortInterceptors(interceptors);
         }
-        return ClientInterceptors.intercept(channel, interceptors);
+        return ClientInterceptors.interceptForward(channel, interceptors);
     }
 
     /**
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
@@ -89,7 +89,8 @@ public abstract class AbstractChannelFactory<T extends ManagedChannelBuilder<T>>
     }
 
     @Override
-    public Channel createChannel(final String name, final List<ClientInterceptor> customInterceptors) {
+    public Channel createChannel(final String name, final List<ClientInterceptor> customInterceptors,
+            boolean sortInterceptors) {
         final Channel channel;
         synchronized (this) {
             if (this.shutdown) {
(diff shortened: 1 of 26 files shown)


ST-DDT:
When this PR is merged, then we should create a new release to publish all those changes.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
```diff
@@ -89,21 +89,20 @@ public final Channel createChannel(final String name) {
...
-        }
-        if (!customInterceptors.isEmpty()) {
-            interceptors.addAll(customInterceptors);
+        final List<ClientInterceptor> interceptors =
+                Lists.newArrayList(this.globalClientInterceptorRegistry.getClientInterceptors());
+        interceptors.addAll(customInterceptors);
+        if (sortInterceptors) {
+            this.globalClientInterceptorRegistry.sortInterceptors(interceptors);
         }
         return ClientInterceptors.intercept(channel, interceptors);
```
yidongnan: Is it also necessary to adjust to `interceptForward` here?

ST-DDT: Yes, you are right. I totally missed that. I will fix it tomorrow.

ST-DDT: Done

PR #241, opened 2019-06-22
Update DiscoveryClientNameResolver implementation

### LINKED ISSUE #240: Random grpc ports, when server restart, client can't update service instance info
DiscoveryClientNameResolver.isNeedToUpdateServiceInstanceList() just only check service's host and port. If server config static server port and random grpc port, the client will not connect to the server.

ST-DDT:
Fixes #240

* Changed the implementation to follow the `DnsNameResolver` one that is provided by gRPC.
* Fixed the server identification incorrectly using the http port instead of the gRPC port

@alienxt Can you test whether this PR fixes the issue for you?

~~I'm not sure whether the current implementation fixes all issues with the discovery client.
Maybe we also have to compare the instance ids as to ensure that gRPC triggers a reconnect to a server that just restarted using the same port again. The automated cloud example test fails here sometimes and I'm not sure whether this is caused by the missing reconnect or just a side effect from the sudden server unavailability.
I will do some more testing, before I consider this ready to be merged.~~

The used eureka/zipkin server used in the example is too slow to always detect/publish the server unavailability / new server in time. That's why it fails sometimes.


ST-DDT:
Testing completed on my side. LGTM

PR #250, opened 2019-07-02
Split AuthenticatingServerInterceptor into an interface

### LINKED ISSUE #249: [Feature Request] Change `AuthenticatingServerInterceptor` to be defined as an interface instead of concrete class in the autoconfiguration class.
`GrpcServerSecurityAutoConfiguration` defines three beans, one of which is:
```
@Bean
@ConditionalOnMissingBean
public AuthenticatingServerInterceptor authenticatingServerInterceptor(final AuthenticationManager authenticationManager, final GrpcAuthenticationReader authenticationReader) {
    return new AuthenticatingServerInterceptor(authenticationManager, authenticationReader);
}
```
Note that `AuthenticatingServerInterceptor` is a concrete class, so if you want to override it, you must do all of the following: (1) exclude `GrpcServerSecurityAutoConfiguration` from auto configuration, (2) define your own configuration class that provides the other two beans and the interceptor that you'd prefer and (3) define your interceptor. 

But if this bean definition were changed to an interface, then all we'd need to do to override it is to define our own bean of that interface type and then the `@ConditionalOnMissingBean` would prevent the default implementation from being added to the `ApplicationContext`.

ST-DDT:
This makes it easier to replace it with custom implementations.
Fixes #249 (@pluttrell)

### DIFF
--- a/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcServerSecurityAutoConfiguration.java
+++ b/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcServerSecurityAutoConfiguration.java
@@ -32,6 +32,7 @@ import net.devh.boot.grpc.server.security.authentication.GrpcAuthenticationReade
 import net.devh.boot.grpc.server.security.check.GrpcSecurityMetadataSource;
 import net.devh.boot.grpc.server.security.interceptors.AuthenticatingServerInterceptor;
 import net.devh.boot.grpc.server.security.interceptors.AuthorizationCheckingServerInterceptor;
+import net.devh.boot.grpc.server.security.interceptors.DefaultAuthenticatingServerInterceptor;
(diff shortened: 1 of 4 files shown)


pluttrell:
Any idea when this might make it into a release?

ST-DDT:
This will be included in the next release unfortunately we cannot give an ETA yet.
I would like to at least fix the bugs referenced in the current milestone.

pluttrell:
The `attachAuthenticationContext()` and `detachAuthenticationContext()` are not applicable to the Reactive streams use case. Including `AbstractAuthenticatingServerCallListener` in the interface implies certain functionality to all implementors, however I have a working `ServerInterceptor` that doesn't use this model and thus this PR is going to still make it tough to override the default implementation provided. 

I'd like to propose that `AuthenticatingServerInterceptor` become simply a marker interface, with perhaps just the `AUTHENTICATION_CONTEXT_KEY` constant. Then the full implementation that was previously the class `AuthenticatingServerInterceptor` be moved to `DefaultAuthenticatingServerInterceptor`.

pluttrell:
LGTM

PR #292, opened 2019-10-28
Log AuthN and AuthZ failures

rkettelerij:
Currently when a client receives an authentication failed or access denied error nothing is logged server-side. This makes debugging extremely difficult. Especially since the exception cause is swallowed.


ST-DDT:
Code wise this looks good to me. Maybe we should move the log statements to the closeCall methods?

I will compare this to spring-security's reference implementation for additional input.

ST-DDT:
It looks like spring-security-web also logs these information so it should be fine to do it as well.
I'll add a few more logging statements.

ST-DDT:
@yidongnan it looks like travis failed to report the build success

PR #312, opened 2019-12-13
Fix memory leak in DiscoveryClientResolverFactory

### LINKED ISSUE #311: DiscoveryClientNameResolver leak and performance issue.
An object leak case has detected recently in out system. grpc-spring-boot-starter version is 2.4.0 and I believe it also exists on master branch. Grpc version is 1.20.0.

Eclipse Memory Analyzer give the following report.

屏幕快照 2019-12-13 14 40 08
屏幕快照 2019-12-13 14 40 22

In method `DiscoveryClientResolverFactory::newNameResolver`,  a `DiscoveryClientNameResolver` is created and add into a list and never removed and cause the leak. 

And things get worse because `DiscoveryClientNameResolver` hold a reference of `Listener` which is a member of `ManagedChannelImpl`, cause `ManagedChannelImpl` leak even if it is shutdown.

A quick fix it that we set listener of `DiscoveryClientNameResolver` to null in method of shutdown and remove `DiscoveryClientNameResolver` from `DiscoveryClientResolverFactory`.

But there is also a performance issue related to `DiscoveryClientNameResolver` when there are many `DiscoveryClientNameResolver`. Each resolver send a request to Eureka to update instance list every 10 seconds even if they are about to resolve the same service. From the above resolver leak case, our Eureka was almost out of service under high workload.

ST-DDT:
Fixes #311 

Use a weak reference collection for storing references in `DiscoveryClientResolverFactory`
Also make sure that we set everything to null on shutdown in `DiscoveryClientResolver`

I will run some basic tests to verify this fix later today, but I'm lacking a project that could reproduce the error.

@wangzw Can you test this fix?


wangzw:
@ST-DDT Thanks for your fast response. #311 actually reports two issues, one is object leak and another is performance issue. This patch only fix the former.

I also works on this issue today. And I'd like to move the work of updating service instance information to `DiscoveryClientResolverFactory` to reduce the workload of Eureka.  If there are a lot of `DiscoveryClientResolver`, it can save a lot.

ST-DDT:
When i wrote the fix, i only assumed that the many entries in that collection caused the performance issue.
Usually the GrpcChannelFactory caches ManagedChannels to the same address (reuse threads and connections) and thus only a very limited number of NameResolvers is created.

Do you have that many targets or is there a different reason why you have name resolvers in your app?

I'm looking forward to your solution.

wangzw:
Our special use case cannot use cached ManagedChannel, we manage channel's life cycle manually. It is why this issue is significant to us. #313 is my attempt of fix.

ST-DDT:
Closing in favor of #313

Mind explaining this special usecase?

wangzw:
```
                                            +
                                           NAT
                                            |
                                       |    |
+------------+       +------------+    |    |     +-----+
| Prometheus |       | Prometheus |    |    |  +--+Agent|
|            +-->+-->+   Proxy    +-+  |    |  |  +-----+
+------------+   |   +------------+ |  |    |  |
                 |                  |  |    |  |
                 |                  <--LB<-----+
                 |                  |  |    |
+------------+   |   +------------+ |  |    |
| Prometheus +-->+---> Prometheus +-+  |    |
|            |       |   Proxy    |    |    |
+------------+       +------------+    |    |
                                       |    |
                                       |    |
                                            |
                                            +

```

We have a monitor system based on prometheus. As you may know that prometheus pull metrics from target system. But the target system is behind the NAT (and something like it) and only has egress connection. 

So we have a prometheus proxy on server side and an agent on target system side to setup a communication tunnel.  Prometheus proxy works as a http proxy and forwards request to target system by agent. Proxy and agent communicate using GRPC Bidirectional Streaming.

Our high available requirement is that the system has no single component failure. So all component (Prometheus, Proxy and LB) works as multiply active instances mode.

The problem is that agent must to connect to all proxies but the proxies are behind the LB. We cannot expose proxies directly to agent but behind the LB by some restrict from customer. So the agent has to constantly try to connect to the proxies var the LB. If the proxy is already connected, disconnect and try again. Because LB is working as not-keep-session mode, agent will eventually connect to all proxy instances. So that prometheus can pull metrics var any proxy on target system by the agent.

If ManagedChannel is cached and reused, agent can only connect to one proxy server forever. That is why we manage the life cycle of channel manually.

Actual system is much complex than above, monitor system is just one piece, a GRPC gateway service (also multiply active instances) is between proxies and LB, it route GRPC request based on method to different backend service and only one GRPC port exposes on LB. The leak happens on gateway service.

ST-DDT:
Thanks for the explanation.

PR #328, opened 2020-02-23
Fix client compatibility with grpc 1.26+

ST-DDT:
This PR updates the dependencies and fixes the client compatibility with grpc classes that are generated using `io.grpc:protoc-gen-grpc-java:1.26.0` or later.

**Tested with**

- 1.25.0
- 1.26.0
- 1.27.1


grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -197,39 +198,81 @@ private GrpcChannelFactory getChannelFactory() {
...
+     * Creates a stub of the given type.
+     *
+     * @param <T> The type of the instance to be injected.
+     * @param stubType The type of the stub to create.
+     * @param channel The channel used to create the stub.
+     * @return The newly created stub.
+     *
+     * @throws BeanInstantiationException If the stub couldn't be created.
+     */
+    protected <T extends AbstractStub<T>> T createStub(final Class<T> stubType, final Channel channel) {
```
mattdkerr: In the future, would it be worth providing a way to use the other signature? (Where they can tweak the CallOptions?)

testExamples.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
```
mattdkerr: FYI I was able to successfully test with this script using Git Bash on Windows.

testExamples.sh
```diff
@@ -11,10 +12,10 @@ localTest() {
 	# Run environment
 	./gradlew :example:local-grpc-server:bootRun -x jar -x classes --console=plain &
 	LOCAL_SERVER=$!
-	sleep 10s
+	sleep 10s # Wait for the server to start
```
mattdkerr: I’ve been using Docker compose and the Gradle plugin’s built-in port check, along with a health check call, to make sure things are running before integration tests start. (Im assuming Kubes with Helm would be the same sort of thing) It’s a bit more deterministic than sleeping but does require docker engine to be running (was a bit tricky on a CI runner).

build.gradle
```diff
@@ -11,29 +11,29 @@ buildscript {
     ext {
         projectVersion = '2.7.0-SNAPSHOT'
 
-        grpcVersion = '1.25.0'
-        protobufVersion = '3.11.0'
-        protobufGradlePluginVersion = '0.8.10'
+        grpcVersion = '1.27.1'
+        protobufVersion = '3.11.4'
+        protobufGradlePluginVersion = '0.8.10' // 0.8.11 causes conflicts (requires gradle update?)
```
mattdkerr: I’ve submitted Issues for this work and will provide PRs soon

ST-DDT:
> Should this be a version bump to 3.0 instead of remaining at 2.7? To me, it’s a break because of gRPC Java breaking things elsewhere potentially.

IMO it's not, because the grpc internal things might have changed, but the overall framework hasn't changed. Do you have any hints/examples for me what could break during the update?
My tests haven't shown any breaking changes with newer grpc versions (server), which is why I didn't detect the incompatibility on the client side n the first place.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
ST-DDT: That's what the `StubTransformer`s are meant for.

See also: https://github.com/yidongnan/grpc-spring-boot-starter/blob/master/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/security/CallCredentialsHelper.java#L117

That way multiple parties could change the stub without affecting each other.

I also thought about using it to inject a custom/mocked stub, but I haven't tested that properly yet and I'm not sure whether that is a good suggestion in the first place.

testExamples.sh
ST-DDT: Thanks for the information. I use it the same way. But on slower computers I had timing issues (especially with the cloud example.). See also my response to the next comment.

testExamples.sh
ST-DDT: I considered just using spring's health endpoint for that and calling it until it it responds with `UP`, but my free time is currently very limited and I would like to avoid having to search for and install additional tools on the Windows PATH just for this simplistic "does the example still start" test.

This would also require the actuator lib in the example, and configuration, ... and I try to keep the examples simple and focused just on grpc. Maybe this would be a nice idea for a full-fledged example that "you can run in production", but right now that isn't on my roadmap.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
ST-DDT: But if you have a suggestion on how to improve this feel free to propose a change or just open an issue for discussion purposes.

ST-DDT:
@yidongnan Could you release this version?

yidongnan:
2.7.0 has been released 😄

ST-DDT:
Oh, I haven't seen it.
Thanks a lot.

PR #351, opened 2020-05-04
Documentation link fixes

NickUfer:
Basically adds to all linked documents the file ending that you can browse the documentation on GitHub.


ST-DDT:
Does github convert them correctly so they still work on the real docs?
https://yidongnan.github.io/grpc-spring-boot-starter/en/

NickUfer:
I merged my changes into the master of my fork and let Github generate a documentation. It seems to work just fine: https://nickufer.github.io/grpc-spring-boot-starter/

But another fact I just found out: The images I thought to be removed are not actually removed. Just removed that second commit.

docs/en/server/security.md
```diff
@@ -277,12 +277,12 @@ public void methodX(Request request, StreamObserver<Response> responseObserver)
...
 ## Additional Topics <!-- omit in toc -->
 
-- [Getting Started](getting-started)
-- [Configuration](configuration)
-- [Contextual Data / Scoped Beans](contextual-data)
-- [Testing the Service](testing)
+- [Getting Started](getting-started.md)
+- [Configuration](configuration.md)
+- [Contextual data / Scoped beans](contextual-data.md)
+- [Testing the service](testing.md)
```
ST-DDT, line 283: Is there a reason why you changed the titles here, but not the actual titles?

NickUfer: No, but it revealed an inconsistency in the documentation capitalisation: https://github.com/yidongnan/grpc-spring-boot-starter/search?q=Contextual+data+%2F+Scoped+beans&unscoped_q=Contextual+data+%2F+Scoped+beans

I copied and replaced most of the changes, that's how it happened. Should I just make it consistent in another commit? If yes which way is the right one?

ST-DDT: > I copied and replaced most of the changes, that's how it happened. Should I just make it consistent in another commit? If yes which way is the right one?

That would be good. IMO we should go for title case -> capitalized words except for articles.

E.g.:
Getting Started
Contextual Data / Scoped Beans
Testing the Service
Configuring a Client

AFAIK the actual page titles should already be like this, so we should check them and use them as is.

Thanks for your help.

PR #368, opened 2020-06-11
fix: The registration conditions of defaultDecompressorRegistry 

ChaselX:
The condition `isForCompression` is the same as CompressorRegistry.  It should be isForDecompression.

change the condition register codec definition on defaultDecompressorRegistry from `isForCompression()` to `isForDecompression()`


yidongnan:
Thanks for fixing these errors

Issue #376, opened 2020-07-06
Request connection on startup


genuss:
**The problem**

I'd like to get immediate feedback on configuration errors via checking a connection to GRPC server on startup.

**The solution**

I propose to modify `net.devh.boot.grpc.client.channelfactory.AbstractChannelFactory#watchConnectivityState()` by making a param passed to `requestConnection` configurable. Configuration param may be set in `GrpcChannelsProperties` preserving default value.

**Alternatives considered**

Another solution could be waiting for the first actual request and later seeing result via `HealthIndicator` but this has drawback for low-load services or forces GRPS-server's developers to implement some kind of health endpoint on their side for the first request.

Unfortunately I can't get a `ManagedChannel` in "user-space" code because it's hidden deep in `InterceptorChannel`

**Additional context**

The best way would be failing a context to start but I couldn't find proper way in GRPC to check for connection in a synchronous way.

I could make a PR, looks simple enough for me.

ST-DDT:
Sounds like a useful feature to me.
IMO we should go for a new property in the config like `grpc.client.<name>.immediateConnect` (feel free to propose a better name).

Requirements:
- The property needs to be false by default
- The description of the property should state, that setting this to true, will only check the connection when the factory is created first and bean creation will fail to be created (thus may be failing the application startup). 
- The startup performance implications should also be stated (potential delay due to waiting for timeout)
- The behavior of the feature should be predictable (grpc's getConnectivityState method needs to be blocking)

> I could make a PR, looks simple enough for me.

We welcome new contributors and I look forward to your contribution. 👍

PR #377, opened 2020-07-06
Request connection on startup

genuss:
Solves #376 

Looks like I was a bit too optimistic about this task. It's not that easy to check that connection was established in GRPC. It doesn't even have such thing as "connection timeout" becase GRPC internally handles connections and retries on failures.

I came out with this solution:
1. Use separate method to check connection on startup if grpc-spring-boot-starter was configured to do so.
2. As far as I understand `ConnectivityState` state-machine it should work like this:
````
IDLE -> CONNECTING -> READY                 for happy path
IDLE -> CONNECTING -> TRANSIENT_FAILURE     for connection failure
````
3. So my method initiates connection via `channel.getState(true)`, registers two callbacks and awaits on `CountDownLatch` for some period of time.
4. If `ConnectivityState` is in wrong status or timeout has passed then we throw exception which stops context from starting.

Well any feedback will be highly appreciated. If you suppose design is OK, I'll make connection timeout configurable too.

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
@@ -30,16 +30,16 @@ import java.util.concurrent.TimeUnit;
 import javax.annotation.PreDestroy;
 import javax.annotation.concurrent.GuardedBy;
 
+import org.springframework.util.unit.DataSize;
+
+import com.google.common.collect.Lists;
+
 import io.grpc.Channel;
 import io.grpc.ClientInterceptor;
 import io.grpc.ClientInterceptors;
 import io.grpc.ConnectivityState;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
-import org.springframework.util.unit.DataSize;
-
-import com.google.common.collect.Lists;
-
 import lombok.extern.slf4j.Slf4j;
 import net.devh.boot.grpc.client.config.GrpcChannelProperties;
 import net.devh.boot.grpc.client.config.GrpcChannelProperties.Security;
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
@@ -19,6 +19,7 @@ package net.devh.boot.grpc.client.channelfactory;
 
 import static java.util.Objects.requireNonNull;
 
+import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -134,16 +135,17 @@ public abstract class AbstractChannelFactory<T extends ManagedChannelBuilder<T>>
         final T builder = newChannelBuilder(name);
         configure(builder, name);
         final ManagedChannel channel = builder.build();
-        if (properties.getChannel(name).isImmediateConnect()) {
-            connectOnStartup(name, channel);
+        final Duration timeout = properties.getChannel(name).getImmediateConnectTimeout();
+        if (!timeout.isZero()) {
+            connectOnStartup(name, channel, timeout);
         }
         watchConnectivityState(name, channel);
         return channel;
     }
 
-    private void connectOnStartup(String name, ManagedChannel channel) {
+    private void connectOnStartup(String name, ManagedChannel channel, Duration timeout) {
         final ConnectivityState state = channel.getState(true);
-        final CountDownLatch timeout = new CountDownLatch(1);
+        final CountDownLatch timeoutLatch = new CountDownLatch(1);
         final AtomicBoolean connected = new AtomicBoolean(false);
 
         channel.notifyWhenStateChanged(state, () -> {
@@ -154,14 +156,14 @@ public abstract class AbstractChannelFactory<T extends ManagedChannelBuilder<T>>
                     ConnectivityState state2 = channel.getState(false);
                     log.info("Second state change {}", state2);
                     connected.set(state2 == ConnectivityState.READY);
-                    timeout.countDown();
+                    timeoutLatch.countDown();
                 });
             }
         });
         boolean timeoutExceeded;
         try {
             log.info("Waiting for connect");
-            timeoutExceeded = !timeout.await(20, TimeUnit.SECONDS);
+            timeoutExceeded = !timeoutLatch.await(timeout.toMillis(), TimeUnit.MILLISECONDS);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             timeoutExceeded = true;
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
@@ -23,7 +23,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.annotation.PreDestroy;
 import javax.annotation.concurrent.GuardedBy;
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java [hunk 1]
```diff
@@ -132,10 +134,43 @@ protected ManagedChannel newManagedChannel(final String name) {
...
+            log.info("First state change {}", state1);
+            if (state1 == ConnectivityState.CONNECTING) {
+                channel.notifyWhenStateChanged(state1, () -> {
+                    ConnectivityState state2 = channel.getState(false);
+                    log.info("Second state change {}", state2);
+                    connected.set(state2 == ConnectivityState.READY);
+                    timeout.countDown();
+                });
+            }
+        });
```
ST-DDT, line 160: How about only checking for `READY` (regardless of the current state)?
Then the timeout would work even if there are temporary failures.

ST-DDT:
> If you suppose design is OK, I'll make connection timeout configurable too.

Maybe we could change the `immediateConnect` to a `immediateConnectTimeout` duration and define it as "do it if > 0"?

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
genuss: If we check for `READY` state only then in case of connection failure we'll have to wait for a full timeout.
If we see that state changed from `CONNECTING` to `TRANSIENT_FAILURE` we may fail faster and user won't have to wait too much.
So except for code beauty first the original approach will be preferable.

genuss:
> Maybe we could change the immediateConnect to a immediateConnectTimeout duration and define it as "do it if > 0"?
That's a great idea. Applied it.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
ST-DDT: OK, then we'll start with this. If feedback requests to change it we might touch it again.

ST-DDT:
Could you please add a test for this feature?
If possible for both code paths:
- Connect success -> app starts
- Connect fails -> app fails to start

I can help you with that, if necessary.

genuss:
I created three tests and arranged them in one file. I think it’s ok to group them so. 
# When connection is not requested, channel must be in IDLE state
# When connection is requested and successful channel must be in READY state when user-code interacts with a channel
# When connection is requested but was unsuccessful, context fails to start. That is the tricky one.

genuss:
Do you have some ideas why build fails? I can't see issues.

ST-DDT:
````txt
ImmediateConnectTest$ImmediateConnectEnabledAndSuccessfulTest > testFailingCall() FAILED
    java.lang.IllegalStateException: Failed to create channel: test
        at net.devh.boot.grpc.client.inject.GrpcClientBeanPostProcessor.processInjectionPoint(GrpcClientBeanPostProcessor.java:126)
		[...]
	Caused by:
        java.lang.IllegalStateException: Can't connect to channel test
            at net.devh.boot.grpc.client.channelfactory.AbstractChannelFactory.connectOnStartup(AbstractChannelFactory.java:172)
            at net.devh.boot.grpc.client.channelfactory.AbstractChannelFactory.newManagedChannel(AbstractChannelFactory.java:140)
````

**Comment by ST-DDT** [comment]
I cannot reproduce the error at my place either. But I assume, that the server might be a bit slow starting up.
Could you do me a favor and change the check to wait for ready? If that does work, we at least know, that it is related to the startup speed.

As an alternative you can also try to uncomment:
````
// failFast = true
// showStandardStreams = true
````
in the main `build.gradle`.

**Comment by genuss** [comment]
This issue with tests was the following. A channel can move to `READY` state too fast (looks like CI on github is), however in my assumption was wrong in the first place. So state diagram can be:
````
IDLE -> CONNECTING -> READY                 for happy path
READY                                       for super-happy path
IDLE -> CONNECTING -> TRANSIENT_FAILURE     for connection failure
````
And I applied your suggestion to wait for ready state. When I actually made it, I understood you're right. The code looks much better now 👍 .

File: grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java [hunk 2]
```diff
@@ -32,12 +34,7 @@
 
 import com.google.common.collect.Lists;
 
-import io.grpc.Channel;
-import io.grpc.ClientInterceptor;
-import io.grpc.ClientInterceptors;
-import io.grpc.ConnectivityState;
-import io.grpc.ManagedChannel;
-import io.grpc.ManagedChannelBuilder;
+import io.grpc.*;
```
ST-DDT, line 37: No star imports please.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
```diff
@@ -254,6 +255,36 @@ protected void watchConnectivityState(final String name, final ManagedChannel ch
...
+        log.info("Successfully connected to channel {}", name);
+    }
+
+    private void waitForReady(ManagedChannel channel, CountDownLatch readySignal) {
+        final ConnectivityState state = channel.getState(false);
+        log.debug("Waiting for ready state. Currently in {}", state);
+        if (state == ConnectivityState.READY) {
+            readySignal.countDown();
+        } else {
+            channel.notifyWhenStateChanged(state, () -> waitForReady(channel, readySignal));
```
ST-DDT, line 284: If the new `state` is failure, does the grpc client retry automatically or would it then wait idling for the timeout?
So maybe add something liek this:

````java
if (state == FAILURE) {
    channel.getState(true);
}
````
(After registering the state change hook)

This could potentially be extended to include the IDLE state, then you don't have to call `channel.getState(true)` in `connectOnStartup` and all (re-)connect would be located in this method.

Comment (ST-DDT) [reply grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java:284]: Yes, you are right. I tested it locally and the client will retry even if he isn't triggered again.

File: tests/src/test/java/net/devh/boot/grpc/test/setup/ImmediateConnectTest.java [hunk 4]
```diff
@@ -0,0 +1,125 @@
...
+public class ImmediateConnectTest {
+
+    @Slf4j
+    @SpringBootTest(properties = {
+            "grpc.client.GLOBAL.address=localhost:9090",
+            "grpc.client.GLOBAL.negotiationType=PLAINTEXT",
+            "grpc.client.GLOBAL.immediateConnectTimeout=10s",
+    })
+    @SpringJUnitConfig(classes = {ServiceConfiguration.class, BaseAutoConfiguration.class})
+    static class ImmediateConnectEnabledAndSuccessfulTest extends AbstractSimpleServerClientTest {
```
ST-DDT, line 47: Please add a `@DirtiesContext` to the class as well, just to be safe. As these errors are really hard to detect (later on).

tests/src/test/java/net/devh/boot/grpc/test/setup/ImmediateConnectTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+import io.grpc.Channel;
+import io.grpc.ConnectivityState;
+import io.grpc.ManagedChannel;
+import lombok.extern.slf4j.Slf4j;
+import net.devh.boot.grpc.client.inject.GrpcClient;
+import net.devh.boot.grpc.test.config.BaseAutoConfiguration;
+import net.devh.boot.grpc.test.config.ServiceConfiguration;
+
+public class ImmediateConnectTest {
```
ST-DDT, line 38: Add a javadoc comment that this test is for GrpcChannelProperties#getImmediateConnectTimeout().

tests/src/test/java/net/devh/boot/grpc/test/setup/ImmediateConnectTest.java
```diff
@@ -0,0 +1,125 @@
...
+                        "grpc.client.GLOBAL.immediateConnectTimeout=1s")
+                .withUserConfiguration(ServiceConfiguration.class, BaseAutoConfiguration.class)
+                .withBean(FailedChannelHolder.class);
+
+        public ImmediateConnectEnabledAndFailedToConnectTest() {
+            log.info("--- ImmediateConnectEnabledAndFailedToConnectTest ---");
+        }
+
+        @Test
+        @DirtiesContext
```
ST-DDT, line 110: AFAICT, this `@DirtiesContext` is useless as the test isn't executed inside a spring context.

tests/src/test/java/net/devh/boot/grpc/test/setup/ImmediateConnectTest.java
ST-DDT: Also rename the class to `ImmediateConnectTests` or similar (otherwise sonar expects a test in this class). The plural name makes it explicit that this class contains more that one test(-setups).

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
genuss: Fixed.

tests/src/test/java/net/devh/boot/grpc/test/setup/ImmediateConnectTest.java
genuss: Fixed.

tests/src/test/java/net/devh/boot/grpc/test/setup/ImmediateConnectTest.java
genuss: Fixed.

tests/src/test/java/net/devh/boot/grpc/test/setup/ImmediateConnectTest.java
genuss: Nice catch! Fixed.

genuss:
Thanks guys! Looking forward to release 🚀

ST-DDT:
FYI: I try to finish the remaining issues in the next two weeks and do a release afterwards.

PR #375, opened 2020-07-05
Support 3rd party gRPC Java compilers via StubFactory

liqweed:
Adds support for 3rd party gRPC compilers like:

- [Reactive gRPC (Reactor)](https://github.com/salesforce/reactive-grpc/tree/master/reactor)
- [Reactive gRPC (RxJava)](https://github.com/salesforce/reactive-grpc/tree/master/rx-java)
- [grpc-kotlin](https://github.com/grpc/grpc-kotlin)
- [ScalaPB](https://scalapb.github.io/grpc.html)
- [akka-grpc](https://github.com/akka/akka-grpc)

Support for server-side services is simple enough by:
- Making the service a Spring bean
- Annotating it with `@GrpcService`

Creating client stubs requires specialized stub factories. This change introduces a `StubFactory` interface:
```
public interface StubFactory {

    AbstractStub<?> createStub(final Class<? extends AbstractStub<?>> stubType, final Channel channel);

    boolean isApplicable(Class<? extends AbstractStub<?>> stubType);
}
```

This allows extending Spring Boot support to 3rd party compiled gRPC stubs by adding beans implementing `StubFactory` like so:
```
@Configuration
public class ReactorGrpcConfiguration {

    @Bean
    StubFactory reactorStubFactory() {
        return new StubFactory() {

            @Override
            public AbstractStub<?> createStub(Class<? extends AbstractStub<?>> stubType, Channel channel) {
                try {
                    final Method factoryMethod = stubType.getDeclaredMethod("newReactorStub", Channel.class);
                    return (AbstractStub<?>) factoryMethod.invoke(null, channel);
                } catch (NoSuchMethodException e) {
                    throw new IllegalArgumentException("Failed to resolve constructor for " + stubType.getName(), e);
                } catch (IllegalAccessException | InvocationTargetException e) {
                    throw new IllegalArgumentException("Failed to instantiate " + stubType.getName(), e);
                }
            }

            @Override
            public boolean isApplicable(Class<? extends AbstractStub<?>> stubType) {
                try {
                    if (stubType.getName().startsWith("Reactor")) {
                        stubType.getDeclaredMethod("newReactorStub", Channel.class);
                        return true;
                    }
                } catch (NoSuchMethodException e) {
                    // not a Reactor gRPC stub
                }
                return false;
            }
        };
    }
}
```
(The code above isn't optimized and serves as an example only)

---
This change is [Reviewable](https://reviewable.io/reviews/yidongnan/grpc-spring-boot-starter/375)

PR #382, opened 2020-07-15
Document StubFactory

### LINKED ISSUE #381: Document StubFactory
Extend the documentation regarding the new `StubFactory` feature.

#375

ST-DDT:
Fixes #381 
Ref: #375 

@liqweed Do you think that this documentation is sufficient or is there something missing?

### DIFF
--- a/docs/en/client/configuration.md
+++ b/docs/en/client/configuration.md
@@ -11,6 +11,7 @@ This section describes how you can configure your grpc-spring-boot-starter clien
 - [Configuration via Beans](#configuration-via-beans)
   - [GrpcChannelConfigurer](#grpcchannelconfigurer)
   - [ClientInterceptor](#clientinterceptor)
+  - [StubFactory](#stubfactory)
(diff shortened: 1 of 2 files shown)


liqweed:
I'm sorry for for the delayed response.
Maybe consider adding a section titled "Support for 3rd party gRPC frameworks" explicitly so it would stand out more for people with this requirement in mind. Also, it's not inherently clear that for the server side of things integration of 3rd party frameworks is a given for any `BindableService` beans. Even though it didn't require any change to the framework, I think it should be explicitly stated. A "Support for 3rd party gRPC frameworks" could be broken into a description both server and client side integration.

ST-DDT:
Are you referring to a top level doc section (same level as trouble-shooting)? Or are you referring to the feature list on the readme?

liqweed:
Both I suppose. I just think it it should be explicitly mentioned (the features list is a good place for that) and documented (either in the top level doc section or in the project's documentation).

PR #412, opened 2020-08-19
Introduces an example project using Kotlin

asarkar:


ST-DDT:
I tried to run the example project using both J8 + J11 and both CMD and Bash, but I always get the following message:

grafic

> 

> ````txt
> $ gradlew.bat clean build --console=plain
> > Task :clean
> > Task :extractIncludeProto
> > Task :extractProto
> > Task :generateProto FAILED
> 
> FAILURE: Build failed with an exception.
> 
> * What went wrong:
> Execution failed for task ':generateProto'.
> > protoc: stdout: . stderr: --grpckt_out: protoc-gen-grpckt: Plugin failed with status code 1.
> 
> 
> * Try:
> Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
> 
> * Get more help at https://help.gradle.org
> 
> BUILD FAILED in 2s
> 4 actionable tasks: 4 executed
> ````

Do I really need a JDK 7 to run it?

ST-DDT:
`grpckt` doesn't seem to be very mature to me.

The above error is related to:
https://github.com/grpc/grpc-kotlin/issues/81#issuecomment-633186948

😞

asarkar:
grpc-KT is not ready for prod. However, I can execute `./gradlew clean build` on the sample project from command line. Are you running differently? I see mentions of `.exe` and JAR launcher in the thread you linked to, but I don’t understand how those are involved if executing Gradle wrapper from command line.

asarkar:
Never mind, I understand that the problem is in their plugin, and not how your run it. It appears on Windows that the plugin expects registry variable to be set for finding the JDK. I don’t have a Windows machine, so this issue didn’t come up for me.

asarkar:
Never mind, I understand that the problem is in their plugin, and not how your run it. It appears on Windows that the plugin expects registry variable to be set for finding the JDK. I don’t have a Windows machine, so this issue didn’t come up for me.

ST-DDT:
Thanks for providing the example, it really helped me reproducing and analyzing the error.
I won't merge it right now, because I thought about integrating it in the main build, but that will probably take some time/attempts.

PR #413, opened 2020-08-19
Add fallback stubfactory

### LINKED ISSUE #410: Stub creation has no fallback
I believe this to be a blocking regression issue.

**The context**

Use [grpc-kotlin](https://github.com/grpc/grpc-kotlin) library to make gRPC calls.

**The bug**

In `grpc-spring-boot-starter:v2.9.0.RELEASE`, `GrpcClientBeanPostProcessor` tries to figure out (reflectively) the appropriate `static` method to create a stub. If it fails, it **falls back** to using the constructor (which is assumed as `private`, but may not be, as in the case of `grpc-kotlin`). In `grpc-spring-boot-starter:v2.10.0.RELEASE`, the fallback is not present, and thus, the stub injection blows up with an error `java.lang.IllegalArgumentException: Unsupported stub type`. See the class hierarchy below:

```
class ServiceNameCoroutineStub @JvmOverloads constructor(
    channel: Channel,
    callOptions: CallOptions = DEFAULT
  ) : AbstractCoroutineStub<ServiceNameCoroutineStub>(channel, callOptions)
```
```
abstract class AbstractCoroutineStub<S: AbstractCoroutineStub<S>>(
  channel: Channel,
  callOptions: CallOptions = CallOptions.DEFAULT
): AbstractStub<S>(channel, callOptions)
```

The problem, that already existed in `grpc-spring-boot-starter:v2.9.0.RELEASE` but was hidden by the fallback, is that there's no code in `GrpcClientBeanPostProcessor` to check for `AbstractStub` subclasses. I think the fallback to using the constructor directly is ok, as seen in the [example code](https://github.com/grpc/grpc-kotlin/blob/master/examples/src/main/kotlin/io/grpc/examples/helloworld/HelloWorldClient.kt#L31), and should not be removed. However, the constructor can be assumed `public` and there's no need to try to change its visibility.

Note that this too, will fail, if grpc-kotlin makes the constructor `private` and adds a static method (I created https://github.com/grpc/grpc-kotlin/issues/163). The futureproof way to fix this is not to check for the stub type, but to check for the existence of static methods named `new*Stub`, and then comparing the return types with the declared stub type.

**Stacktrace and logs**

```
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.mycompany.ServiceNameGrpcKt$ServiceNameCoroutineStub]: Unsupported stub type: com.mycompany.ServiceNameGrpcKt$ServiceNameCoroutineStub -> Please report this issue.
	at net.devh.boot.grpc.client.inject.GrpcClientBeanPostProcessor.lambda$createStub$1(GrpcClientBeanPostProcessor.java:244)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at net.devh.boot.grpc.client.inject.GrpcClientBeanPostProcessor.createStub(GrpcClientBeanPostProcessor.java:243)
	at net.devh.boot.grpc.client.inject.GrpcClientBeanPostProcessor.valueForMember(GrpcClientBeanPostProcessor.java:218)
	at net.devh.boot.grpc.client.inject.GrpcClientBeanPostProcessor.processInjectionPoint(GrpcClientBeanPostProcessor.java:127)
	at net.devh.boot.grpc.client.inject.GrpcClientBeanPostProcessor.postProcessBeforeInitialization(GrpcClientBeanPostProcessor.java:83)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.applyBeanPostProcessorsBeforeInitialization(AbstractAutowireCapableBeanFactory.java:416)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1788)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:595)
```

**Steps to Reproduce**

N/A

Which versions do you use?
* Spring (boot): 2.3.1.RELEASE
* grpc-java: 1.30.2
* grpc-spring-boot-starter: 2.9.0.RELEASE

**Additional context**

* Did it ever work before? Yes.
* Do you have a demo? No.

ST-DDT:
Fixes #410 

@asarkar I used the kotlin project to test this fix and it seems to be working as expected. Could you please try and confirm that it solves the problem for you as well?

- [x] Implement fix
- [x] Manually confirm that it "works for me"
- [x] Manually confirm that it "works for the issue reporter"
- [x] Add automated tests for it


asarkar:
> Could you please try

Can you please upload a JAR here that you built locally? I don't want to go through cloning the repo and building it locally.

ST-DDT:
> > Could you please try
> 
> Can you please upload a JAR here that you built locally? I don't want to go through cloning the repo and building it locally.

Sure [grpc-client-spring-boot-autoconfigure 2.10.1-SNAPSHOT.zip](https://github.com/yidongnan/grpc-spring-boot-starter/files/5101194/2.10.1-SNAPSHOT.zip)

asarkar:
Seems to be working with the example project. I see that you have left the other `StubFactory` implementation classes; isn't the `FallbackStubFactory` going to replace them all? It probably should be called `DefaultStubFactory`.

ST-DDT:
> Seems to be working with the example project. I see that you have left the other `StubFactory` implementation classes; isn't the `FallbackStubFactory` going to replace them all? It probably should be called `DefaultStubFactory`.

I have issues with managing the order of them. Setting `@Order(HIGHEST/LOWEST)` doesn't seem to have any impact on them.
Thats why I kept them as is for now.

Also I'm not sure whether I should cache the factory methods/constructors for subsequent calls.

yidongnan (approved):
LGTM

yidongnan:
@ST-DDT 2.10.1 just released

PR #417, opened 2020-08-22
Add support for @GrpcService annotations in @Configuration

### LINKED ISSUE #406: Allow GrpcService on @Bean methods
**The problem**

`GrpcService` is only applicable to types.

**The solution**

Allow `GrpcService` on methods, so that it can be used in `@Configuration` classes. This allows for using non-Spring/third-party `BindableService` in Spring gRPC code.

**Alternatives considered**

A custom `GrpcServiceDiscoverer` can be created, but that requires more code, defeats the purpose of autoconfiguration, and disables `AnnotationGrpcServiceDiscoverer`. I think the last assumption is incorrect; there's no reason we can't have one class annotated with `GrpcService`, and another `@Configuration` method with annotated with `GrpcService`.

**Additional context**

Spring Boot 2.3.1.RELEASE, grpc-spring-boot-starter 2.9.0.RELEASE

ST-DDT:
Fixes #406

Example:

````java
@Configuration
public class BeanAnnotatedServiceConfig {

    @GrpcService
    SomeServiceImpl someServiceImpl() {
        return new SomeServiceImpl();
    }

}
````

### DIFF
--- a/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/service/GrpcService.java
+++ b/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/service/GrpcService.java
@@ -23,6 +23,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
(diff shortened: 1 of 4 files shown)


yidongnan (approved):
LGTM

PR #435, opened 2020-10-02
Improve NameResolverRegistry registration logs

### LINKED ISSUE #423: Log NameResolverProvider class, not instance
**The problem**
```
2020-09-04 00:25:31.795 [Test worker] [INFO ] n.d.b.g.c.n.NameResolverRegistration - net.devh.boot.grpc.server.nameresolver.SelfNameResolverFactory@2bc9c324 is available -> Added to the NameResolverRegistry
```

```
net.devh.boot.grpc.server.nameresolver.SelfNameResolverFactory@138b9abe was removed from the NameResolverRegistry 
```

**The solution**

These message are not really useful. It seems the `NameResolverProvider` object id is logged, whereas the class name would be more useful. Also, such lifecycle message are probably better fit for `DEBUG` level, not `INFO` level.

**Alternatives considered**

None.

ST-DDT:
Fixes #423 

- Add SelfNameResolverFactory#toString()
- Use debug log levels

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ buildscript {
         }
     }
     ext {
-        projectVersion = '2.10.1.RELEASE'
+        projectVersion = '2.11.0-SNAPSHOT'
 
         // https://github.com/grpc/grpc-java/releases
         grpcVersion = '1.31.1'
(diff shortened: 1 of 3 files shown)


yidongnan (approved):
LGTM

PR #454, opened 2020-11-17
Added service grace off support

klboke:
Add graceful shutdown support. This is a problem found in our online environment. Our online traffic is relatively large. During each rolling update, because the service is shut down too quickly, some requests are not completed, causing a large number of requests to report errors. When the main thread is closed, block and wait for all threads to be processed to solve the problem

### DIFF
--- a/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/serverfactory/GrpcServerLifecycle.java
+++ b/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/serverfactory/GrpcServerLifecycle.java
@@ -112,6 +112,11 @@ public class GrpcServerLifecycle implements SmartLifecycle {
         final Server localServer = this.server;
         if (localServer != null) {
             localServer.shutdown();
+            try {
+                this.server.awaitTermination();
+            } catch (final InterruptedException e) {
+                Thread.currentThread().interrupt();
+            }
             this.server = null;
             log.info("gRPC server shutdown.");
         }


ST-DDT:
Please add a comment to the block why you do it there in addition to the other await block:
https://github.com/yidongnan/grpc-spring-boot-starter/blob/bf5c22217b7e38d245ded0ae255f0b4bd77873cf/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/serverfactory/GrpcServerLifecycle.java#L94-L100

I assume that other required beans are shutdown before the last request completes, thus causing the issues. So blocking the "shutdown" thread until grpc-server shutdown completes prevents this issue. Is my assumption correct?

Have you tested your change locally? These scenarios are very difficult to reproduce (and I have never had this problem myself), so I trust your word if it solves the problem for you.

klboke:
> Please add a comment to the block why you do it there in addition to the other await block:
> https://github.com/yidongnan/grpc-spring-boot-starter/blob/bf5c22217b7e38d245ded0ae255f0b4bd77873cf/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/serverfactory/GrpcServerLifecycle.java#L94-L100
> 
> I assume that other required beans are shutdown before the last request completes, thus causing the issues. So blocking the "shutdown" thread until grpc-server shutdown completes prevents this issue. Is my assumption correct?
> 
> Have you tested your change locally? These scenarios are very difficult to reproduce (and I have never had this problem myself), so I trust your word if it solves the problem for you.
Yes, we have fully verified it, and this is very easy to verify. Simulate business blocking in the grpc service for 60 seconds, and then kill the service. When server.awaitTermination() is not added, the service will be closed immediately. After server.awaitTermination(), it will wait for the service to complete before closing. This server.awaitTermination() must be added to the main thread to avoid immediate death of the main thread.

klboke:
Yes, we have fully verified it, and this is very easy to verify. Simulate business blocking in the grpc service for 60 seconds, and then kill the service. When server.awaitTermination() is not added, the service will be closed immediately. After server.awaitTermination(), it will wait for the service to complete before closing. This server.awaitTermination() must be added to the main thread to avoid immediate death of the main thread.

ST-DDT:
I considered this some more and I'm no longer sure where the exact issue is.
IMO it does not matter whether the main thread is alive or not.
The 60 seconds you mentioned in your post are most likely related to the request timeout.
We don't invoke `server.shutdownNow()` so the server does not stop the processing of the actual request until it is completed.
The `grpc-server-container` Thread is explicitly a non-daemon thread so it should keep the JVM alive (it's its only purpose).

* Please provide a test that can be used to reproduce the issue and can be used to verify that we won't re-introduce the bug accidentally in the future. As an alternative please provide a stacktrace that helps me to understand the underlying issue.
* Also as mentioned above, please add a comment to your block, that explains why it is there to future readers of the code.

ST-DDT (changes requested):
I considered this some more and I'm no longer sure where the exact issue is.
IMO it does not matter whether the main thread is alive or not.
The 60 seconds you mentioned in your post are most likely related to the request timeout.
We don't invoke server.shutdownNow() so the server does not stop the processing of the actual request until it is completed.
The grpc-server-container Thread is explicitly a non-daemon thread so it should keep the JVM alive (it's its only purpose).

* Please provide a test that can be used to reproduce the issue and can be used to verify that we won't re-introduce the bug accidentally in the future. As an alternative please provide a stacktrace that helps me to understand the underlying issue.
* Also as mentioned above, please add a comment to your block, that explains why it is there to future readers of the code.

klboke:
Yes, although shutdown() is used and shutdownNow() is not used, there are still problems. The documentation of the shutdown() method describes that after calling the shutdown() method, new requests will be rejected, and the processing requests will continue. But it will not wait for the processing to complete, you must use awaitTermination() to wait for the call to complete
```java
  /**
   * Initiates an orderly shutdown in which preexisting calls continue but new calls are rejected.
   * After this call returns, this server has released the listening socket(s) and may be reused by
   * another server.
   *
   * <p>Note that this method will not wait for preexisting calls to finish before returning.
   * {@link #awaitTermination()} or {@link #awaitTermination(long, TimeUnit)} needs to be called to
   * wait for existing calls to finish.
   *
   * @return {@code this} object
   * @since 1.0.0
   */
  public abstract Server shutdown();
```

klboke:
> I considered this some more and I'm no longer sure where the exact issue is.
> IMO it does not matter whether the main thread is alive or not.
> The 60 seconds you mentioned in your post are most likely related to the request timeout.
> We don't invoke server.shutdownNow() so the server does not stop the processing of the actual request until it is completed.
> The grpc-server-container Thread is explicitly a non-daemon thread so it should keep the JVM alive (it's its only purpose).
> 
> * Please provide a test that can be used to reproduce the issue and can be used to verify that we won't re-introduce the bug accidentally in the future. As an alternative please provide a stacktrace that helps me to understand the underlying issue.
> * Also as mentioned above, please add a comment to your block, that explains why it is there to future readers of the code.

Refer to the blog post record for the verification process：http://www.kailing.pub/article/index/arcid/314.html

cyqlegend:
> I considered this some more and I'm no longer sure where the exact issue is.
> IMO it does not matter whether the main thread is alive or not.
> The 60 seconds you mentioned in your post are most likely related to the request timeout.
> We don't invoke server.shutdownNow() so the server does not stop the processing of the actual request until it is completed.
> The grpc-server-container Thread is explicitly a non-daemon thread so it should keep the JVM alive (it's its only purpose).
> 
> * Please provide a test that can be used to reproduce the issue and can be used to verify that we won't re-introduce the bug accidentally in the future. As an alternative please provide a stacktrace that helps me to understand the underlying issue.
> * Also as mentioned above, please add a comment to your block, that explains why it is there to future readers of the code.

I tested  awaitTermination() and it works as expect. you may refer to https://grpc.github.io/grpc-java/javadoc/io/grpc/Server.html#awaitTermination--
Note that this method (shutdown) will not wait for preexisting calls to finish before returning. awaitTermination() or awaitTermination(long, TimeUnit) needs to be called to wait for existing calls to finish.

ST-DDT:
Sorry for the delay, but I'm really busy right now, I will report back to you once I have some time to look into it in more depth.

ST-DDT:
Thanks for your contribution. I couldn't update this PR for some reason, so I started a new one. I added a configurable shutdownGracePeriod in #491 based on a few comments in issues that I lost track of.

PR #485, opened 2021-01-21
exception handling - best match

anjeyy:
In Addition to #465 , a more precise solution to resolving exception. Now Supports a convenient way of handling inherited Exceptions.

---

Example:

The following Structure of Exceptions is given and a **FirstLevelException** is thrown inside your application code
```java
public class ExceptionContainer {
    public static class MyRootRuntimeException extends RuntimeException {
        public MyRootRuntimeException(String msg) {
            super(msg);
        }
    }

    public static class FirstLevelException extends MyRootRuntimeException {
        public FirstLevelException(String msg) {
            super(msg);
        }
    }
}
```

Now with the following Advice you will get a precise Exception Handling where a thrown FirstLevelException is correctly executed for the precise depth of mapping. Without this PR, the execution **can** be for mapped _MyRootRuntimeException_.
```java
@GrpcAdvice
public class GrpcExceptionAdvice {

    @GrpcExceptionHandler
    public Status handleInvalidArgument(MyRootRuntimeException e) {
        return Status.INVALID_ARGUMENT.withDescription("Your description").withCause(e);
    }

    @GrpcExceptionHandler(FirstLevelException.class)
    public StatusException handleResourceNotFoundException(FirstLevelException e) {
        Status status = Status.NOT_FOUND.withDescription("Your description").withCause(e);
        return status.asException();
    }

}
```

---

TODO
- [x] implementation ✔️ 
- [x] unit tests ✔️ 
- [x] review ✔️

### DIFF
--- a/tests/src/test/java/net/devh/boot/grpc/test/advice/AdviceExceptionHandlingTest.java
+++ b/tests/src/test/java/net/devh/boot/grpc/test/advice/AdviceExceptionHandlingTest.java
@@ -29,6 +29,7 @@ import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.ImportAutoConfiguration;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.security.authentication.AccountExpiredException;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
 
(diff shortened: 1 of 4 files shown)


tests/src/test/java/net/devh/boot/grpc/test/advice/AdviceExceptionHandlingTest.java
```diff
@@ -119,7 +121,8 @@ void testThrownClassCastException_IsMappedAsStatusRuntimeExceptionAndWithMetadat
     @DirtiesContext
     void testThrownMyRootRuntimeException_IsNotMappedAndResultsInInvocationException() {
```
ST-DDT, line 122: This test name should be adjusted as well.

ST-DDT (changes requested):
One little change otherwise it looks good to me.

tests/src/test/java/net/devh/boot/grpc/test/advice/AdviceExceptionHandlingTest.java
anjeyy: You're right. Done.

ST-DDT (approved):

ST-DDT:
Thanks for the improvement.

tests/src/test/java/net/devh/boot/grpc/test/advice/AdviceExceptionHandlingTest.java
ZMRWEGO: yes is‘s right

ZMRWEGO: test

Issue #433, opened 2020-09-28
Provide a gRPC server health check


asarkar:
**The problem**

Monitor the health of a gRPC server app.

**The solution**

Provide a gRPC server `HealthEndpoint`. My blog post on this matter may be useful.
https://blog.asarkar.com/technical/grpc-kubernetes-spring/

**Alternatives considered**

Implement myself.

**Additional context**

Also note that starting with version 2.3.0.RELEASE, Spring Boot provides liveness and readiness information under Actuator health endpoint. See [Kubernetes Probes](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-features.html#production-ready-kubernetes-probes) for details.

ST-DDT:
I'm not sure what you are exactly asking for.
* Export some kind of `HealthIndicator` for the grpc-server to spring actuator,
* or use the `HealthIndicator`s to populate a `grpcHealthEndpoint` call/service that mirrors the web endpoint,
* or use the `HealthIndicator`s to populate the `HealthStatusManager`

asarkar:
> Export some kind of HealthIndicator for the grpc-server to spring actuator

This. I've a Kotlin implementation that I can share, which I'm sure can be easily turned into Java.

ST-DDT:
I'm not aware of any useful grpc-server availability indicator, so I'm very interested in your implementation. Can you link it here?
I will decide, once I have seen the implementation.

asarkar:
```
open class GrpcServerHealthIndicator internal constructor(private val healthChannel: ManagedChannel) : HealthIndicator {
    internal constructor(port: Int, channelBuilderClass: String) : this(newChannel(port, channelBuilderClass))

    private val healthStub: HealthGrpc.HealthBlockingStub = HealthGrpc.newBlockingStub(healthChannel)

    private var availabilityChangeEventPublishMethod: Method? = null
    private var publishAvailabilityChangeEvent: Boolean = true
    private var readinessStateRefusingTraffic: Enum<*>? = null
    private var livenessStateBroken: Enum<*>? = null

    @Autowired
    lateinit var context: ApplicationContext

    @PostConstruct
    open fun postConstruct() {
        publishAvailabilityChangeEvent = context.containsBean("livenessStateHealthIndicator") &&
                context.containsBean("readinessStateHealthIndicator")
        if (publishAvailabilityChangeEvent) {
            log.info("Kubernetes probes are enabled")
        } else {
[...]

Issue #411, opened 2020-08-18
java.lang.ClassNotFoundException: io.grpc.services.HealthStatusManager


asarkar:
**The context**

Create a gRPC server using `grpc-spring-boot-starter` library.

**The bug**

Even when `grpc.server.healthServiceEnabled=false`, `GrpcServerAutoConfiguration` tries to create a `HealthStatusManager`, which fails with a `ClassNotFoundException` if `io.grpc:grpc-services` is not on the classpath. Now, if the application has indicated that it doesn't want to enable health service, there's no reason for it to add an unnecessary jar on the classpath that is "secretly" needed by the `grpc-spring-boot-starter` library.

My suggestion is to check for the presence of the class, as well as the property `grpc.server.healthServiceEnabled`, before creating the `HealthStatusManager`.

**Stacktrace and logs**

```
java.lang.IllegalStateException: Error processing condition on net.devh.boot.grpc.server.autoconfigure.GrpcServerAutoConfiguration.defaultGrpcServerProperties
	at org.springframework.boot.autoconfigure.condition.SpringBootCondition.matches(SpringBootCondition.java:60)
	at org.springframework.context.annotation.ConditionEvaluator.shouldSkip(ConditionEvaluator.java:108)
	...
Caused by: java.lang.IllegalStateException: Failed to introspect Class [net.devh.boot.grpc.server.autoconfigure.GrpcServerAutoConfiguration] from ClassLoader [jdk.internal.loader.ClassLoaders$AppClassLoader@9e89d68]
	at org.springframework.util.ReflectionUtils.getDeclaredMethods(ReflectionUtils.java:481)
	...
Caused by: java.lang.NoClassDefFoundError: io/grpc/services/HealthStatusManager
	at java.base/java.lang.Class.getDeclaredMethods0(Native Method)
	...
Caused by: java.lang.ClassNotFoundException: io.grpc.services.HealthStatusManager
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	...
```

**Steps to Reproduce**

N/A

**The application's environment**

Which versions do you use?
* Spring (boot): 2.3.1.RELEASE
* grpc-java: 1.30.2
* grpc-spring-boot-starter: 2.9.0.RELEASE

**Additional context**

* Did it ever work before? No
* Do you have a demo? No

ST-DDT:
If I have enough time I will fix it in 1.10.1

[...]

PR #488, opened 2021-01-25
Ability to expose gRPC health through actuator

eleduardo:
Implements #433 

This PR implements a reactive healthcheck (made it reactive since exposing actuator via webflux is common) that exposes the status of the gRPC server.

The healthcheck is perfromed hitting the actual service since that actually proves that the server is indeed responding. It also follows a strict (yet configurable) deadline and forces a non-retry.

I did not send the event to the liveness/readiness since the whole health endpoint goes to 503 if a component is marked as out of traffic so /actuator/health can be used directly but nothing prevents the bean from also sending the event to the readiness/liveness probe.

No specific tests added in this PR or docs but pending comments I can make changes.

### DIFF
--- a/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/health/GrpcServerActuatorHealthIndicator.java
+++ b/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/health/GrpcServerActuatorHealthIndicator.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2016-2021 Michael Zhang <yidongnan@gmail.com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+ * documentation files (the "Software"), to deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+ * WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package net.devh.boot.grpc.server.health;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.springframework.boot.actuate.health.AbstractReactiveHealthIndicator;
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.actuate.health.Status;
+
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+
+import io.grpc.Channel;
+import io.grpc.Deadline;
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.health.v1.HealthCheckRequest;
+import io.grpc.health.v1.HealthCheckResponse;
+import io.grpc.health.v1.HealthGrpc;
+import reactor.core.publisher.Mono;
+
+/**
+ * This class will expose the status of the server into the actuator when it has the "show-components" configuration
+ * set. To retrieve the status the health probe actually hits the server and maps the status
+ */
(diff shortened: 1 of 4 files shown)


eleduardo:
Hi @ST-DDT or @yidongnan is there any chance to get feedback on this PR?

ST-DDT:
Yes, sorry for the delay. I try to wrap up the next release (v2.11.0) and currently don't have time to add a new feature to it. That being said, this won't be in the next release.

As for the actual content. You implementation uses reactive code, which isn't bad in itself, but it forces the users to add it to their code as well as seen in the tests. So maybe (also) adding a non-reactive variant might be a good alternative.
There is also https://github.com/yidongnan/grpc-spring-boot-starter/issues/411, which tries to run without the health service dependency. So I considered removing it from the default stack entirely/making it optional (especially if it does not serve any health states by default). A bridge might still be useful though.

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/health/GrpcServerActuatorHealthIndicator.java
```diff
@@ -0,0 +1,100 @@
...
+    public GrpcServerActuatorHealthIndicator(String host, int port, long deadline) {
+        String internalHost = host.equals("*") ? "0.0.0.0" : host;
+        Channel channel = ManagedChannelBuilder
+                .forAddress(internalHost, port)
+                .disableRetry()
+                .build();
+        healthCheckClient = HealthGrpc.newFutureStub(channel);
+        if (deadline > 0) {
+            this.deadlineMs = deadline;
+        }
```
ST-DDT, line 61: IMO this should fail, if the deadline is invalid.

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/health/GrpcServerActuatorHealthIndicator.java
```diff
@@ -0,0 +1,100 @@
...
+    private HealthGrpc.HealthFutureStub healthCheckClient;
+    private long deadlineMs = 500;
+
+    public GrpcServerActuatorHealthIndicator(String host, int port, long deadline) {
+        String internalHost = host.equals("*") ? "0.0.0.0" : host;
+        Channel channel = ManagedChannelBuilder
+                .forAddress(internalHost, port)
+                .disableRetry()
+                .build();
+        healthCheckClient = HealthGrpc.newFutureStub(channel);
```
ST-DDT, line 58: There should be a constructor that passes this stub directly.

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/health/GrpcServerActuatorHealthIndicator.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * This class will expose the status of the server into the actuator when it has the "show-components" configuration
+ * set. To retrieve the status the health probe actually hits the server and maps the status
+ */
+public class GrpcServerActuatorHealthIndicator extends AbstractReactiveHealthIndicator {
+
+    private final static String STATUS = "status";
+    private final static String DOWN_STATUS = "DOWN";
+    private HealthGrpc.HealthFutureStub healthCheckClient;
```
ST-DDT, line 49: IMO a reactive class should use a reactive stub as well.
(I'm not sure whether there is a reactive implementation for this stub though)

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/health/GrpcServerActuatorHealthIndicator.java
```diff
@@ -0,0 +1,100 @@
...
+                    if (HealthCheckResponse.ServingStatus.SERVING.equals(status)) {
+                        builder.withDetail(STATUS, status.toString()).status(Status.UP);
+                    } else if (HealthCheckResponse.ServingStatus.NOT_SERVING.equals(status)) {
+                        builder.withDetail(STATUS, status.toString()).status(Status.OUT_OF_SERVICE);
+                    } else {
+                        builder.withDetail(STATUS, status.toString()).status(Status.UNKNOWN);
+                    }
+                    return builder.build();
+                })
+                .onErrorReturn(builder.withDetail(STATUS, DOWN_STATUS).status(Status.DOWN).build());
```
ST-DDT, line 81: I'm not sure whether the `withDetail(STATUS, *)` serves any additional information in addition to `status(DOWN)` here.

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/health/GrpcServerActuatorHealthIndicator.java
```diff
@@ -0,0 +1,100 @@
...
+        if (deadline > 0) {
+            this.deadlineMs = deadline;
+        }
+    }
+
+    @Override
+    protected Mono<Health> doHealthCheck(Health.Builder builder) {
+        ListenableFuture<HealthCheckResponse> responseFuture = healthCheckClient
+                .withDeadline(Deadline.after(deadlineMs, TimeUnit.MILLISECONDS))
+                .check(HealthCheckRequest.newBuilder().build());
```
ST-DDT, line 68: What about the other `services`?

ST-DDT (changes requested):

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/health/GrpcServerActuatorHealthIndicator.java
eleduardo: @ST-DDT I need to take a look at how spring does it but wondering if it is really worth having both... I'll double check and if the actuator running on a reactive model protects itself from the blocking implementation I'll just implement it with blocking code

ST-DDT:
Closing because requested changes haven't been made/addressed.

Issue #396, opened 2020-07-30
Introduce configuration option for force shutdown client/server


asarkar:
In a regular application without using this library, we would do something like this in the client code to shutdown the `ManagedChannel`:
```
channel
  .shutdown()
  .awaitTermination(5, TimeUnit.SECONDS)
```

 and on the serve side:
```
server.shutdown()
  .awaitTermination(5, TimeUnit.SECONDS)
```

What is the recommended way to release resources when using this library?

ST-DDT:
This library takes care of it:

https://github.com/yidongnan/grpc-spring-boot-starter/blob/master/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java#L300

https://github.com/yidongnan/grpc-spring-boot-starter/blob/master/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/serverfactory/GrpcServerLifecycle.java#L111

asarkar:
> If the channels don't have terminated after 60 seconds then they will be forcefully shutdown

In my integration tests, I'm noticing a long time for them to finish _after_ my test code is executed. Could this be because of the above?

It seems the wait time is neither configurable, nor avoidable. In tests, I might not want to wait as long as 60 seconds. What do you think about introducing two new parameters for the client, `shutdownWaitTime` and `shutdownNow`? In the second case, `shutdownWaitTime` would be zero, no matter what the user set it to.

ST-DDT:
> > If the channels don't have terminated after 60 seconds then they will be forcefully shutdown
> 
> In my integration tests, I'm noticing a long time for them to finish _after_ my test code is executed. Could this be because of the above?

Do you see a lot of `Awaiting channel shutdown:` message in your log?

If yes, then there is still a call running on that channel.
If not, then you might not cause spring/the channel factory to shutdown. Try `@DirtiesContext`

> It seems the wait time is neither configurable, nor avoidable. In tests, I might not want to wait as long as 60 seconds. What do you think about introducing two new parameters for the client, shutdownWaitTime and shutdownNow? In the second case, shutdownWaitTime would be zero, no matter what the user sets it to.

I agree, that such a config option might be desirable. However, I think a single property should be enough. Because if it is set to `0` it will shutdown now.

> The same settings could be applied to the server for consistency.

I agree.

asarkar:
[...]

PR #491, opened 2021-01-31
Wait for the grpc-server to gracefully shutdown

ST-DDT:
Supersedes #454 (Based on @klboke's PR)
Adds #396

Otherwise the spring context might already be shutdown and the calls
fail, because other components are already shutdown.

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/channelfactory/AbstractChannelFactory.java
@@ -17,12 +17,16 @@
 
 package net.devh.boot.grpc.client.channelfactory;
 
+import static java.util.Comparator.comparingLong;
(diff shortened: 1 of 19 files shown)


ST-DDT:
I'm not sure which term/property is better:
`gracefullShutdownTimeout` or `shutdownGracePeriod`

> `server.shutdown.grace-period=30s` (Spring-Web)

Only after completing the PR did I think of the later. What do you think?

asarkar:
There seems to be multiple instances of this comment:
> If set to {@code -1} the server waits forever

How does one set a `Duration` to `-1`?

ST-DDT:
@asarkar Thanks for the hint. I adjusted the wording there.
All negative values will result in an infinite timeout.
You can set them like this: `grpc.server.shutdownGracePeriod=-1` (with or without unit suffix)

yidongnan:
> I'm not sure which term/property is better:
> `gracefullShutdownTimeout` or `shutdownGracePeriod`
> 
> >

@ST-DDT shutdownGracePeriod looks better.

yidongnan (approved):

PR #494, opened 2021-02-05
Update client configuration metadata

LeoBarykin:
I was looking for an option to init connection on application startup. This configuration fits perfectly, but I have spent quite a time, since all other options are well described in additional-spring-metadata.


ST-DDT (approved):

ST-DDT:
I fixed a small typo in the property name and added also another missing one.
Thanks for your contribution and bringing this to our attention. 👍

Issue #495, opened 2021-02-07
Tags vs Metadata Consul Service Discovery


tentativafc:
**Error when read tag gRPC.port from Consul Spring Cloud Discovery**

**The bug**

I cannot put the **gRPC.port** consul metadada, because consul does not support dot character, and the client does not read Tags.

**spring.cloud.consul.discovery.metadata.gRPC.port=9090**

Invalid Service Meta: Couldn't load metadata pair ('gRPC.port', '9090'): Key contains invalid characters

Reading the link bellow i found this warning:

https://docs.spring.io/spring-cloud-consul/docs/current/reference/html/index.html
...Older versions of Spring Cloud Consul populated the ServiceInstance.getMetadata() method from Spring Cloud Commons by parsing the spring.cloud.consul.discovery.tags property. This is no longer supported, please migrate to using the spring.cloud.consul.discovery.metadata map....

**Steps to Reproduce**

1. Run Consul server discovery
2. Start GRPC server
3. Configure GRPC client using discovery schema. ex: grpc.client.data-service.address=discovery:///data-service
4. Put breakpoint on https://github.com/yidongnan/grpc-spring-boot-starter/blob/master/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java ,  method **getGRPCPort**
```java
  private int getGRPCPort(final ServiceInstance instance) {
            final Map<String, String> metadata = instance.getMetadata();
            if (metadata == null) {
                return instance.getPort();
            }
            final String portString = metadata.get(GrpcUtils.CLOUD_DISCOVERY_METADATA_PORT);
            if (portString == null) {
                return instance.getPort();
            }
            try {
                return Integer.parseInt(portString);
            } catch (final NumberFormatException e) {
                // TODO: How to handle this case?
                throw new IllegalArgumentException("Failed to parse gRPC port information from: " + instance, e);
            }
        }
```
**The application's environment**

https://github.com/tentativafc/poc-orchestration-microservice/commits/service_discovery
commit: ba438aa98c7c50f27cf02605faf76a8a6170cf06

Which versions do you use?
* Spring (boot): 2.4.2
* Srping Cloud: 2020.0.1
* grpc-spring-boot-starter: 2.10.1.RELEASE
* java: version + architecture: OpenJDK 64-Bit Server VM (build 11.0.9.1+1-Ubuntu-0ubuntu1.20.10)

PR #496, opened 2021-02-07
Change metadata from gRPC.port to gRPC_port

tentativafc:
Fix this bug: https://github.com/yidongnan/grpc-spring-boot-starter/issues/495


build.gradle
```diff
@@ -9,7 +9,7 @@ buildscript {
         }
     }
     ext {
-        projectVersion = '2.11.0-SNAPSHOT'
+        projectVersion = '2.11.1-SNAPSHOT'
```
ST-DDT, line 12: There is no need to bump the version as it is already a snapshot.

ST-DDT:
We should at least maintain backwards compatibility in the discovery client resolver by also checking the dot variant.

@yidongnan what are your thoughts on changing the tag name?

ST-DDT:
This issue seems not to be present in spring `v2.3.x` (the used library versions) => I'm going to delay this to the next release.

yidongnan:
> We should at least maintain backwards compatibility in the discovery client resolver by also checking the dot variant.
> 
> @yidongnan what are your thoughts on changing the tag name?

I think we should provide the failback logic. If `grpc_port` is empty, we try to use `grpc.port` and provide the warning log, This can reduce the impact and urge users to upgrade. In the next version, we can remove this failback logic.

ST-DDT (approved):

yidongnan (approved):

PR #509, opened 2021-03-15
docs(settings): add info about available settings for configuration via application.yaml

pumano:


ST-DDT:
IMO this section is really similar to the one at the top (L25++):
https://yidongnan.github.io/grpc-spring-boot-starter/en/server/configuration.html#configuration-via-properties
Is there something about that section that needs improvement, is unclear, or is maybe in bad order?

pumano:
@ST-DDT looks like I don't look at it properly. Because it's clickable links (to javadoc or classes) not only text examples. So, I close it.

PR #512, opened 2021-03-19
Improve build to benefit from incremental builds

totorean:
1. Enable caching, parallel execution, file system watching and the gradle daemon in `gradle.properties` file. For a better understanding of what each does see [gradle configuration properties](https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties)
2. Ignore the `'Build-Time'` attribute in manifest files that is always changing. See [normalization](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:meta_inf_normalization)
3. Exclude all build folders and files from spotless

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,7 @@ plugins {
     id 'com.google.protobuf' version '0.8.14'
     id 'io.franzbecker.gradle-lombok' version '4.0.0' apply false
     id 'com.github.ben-manes.versions' version '0.36.0' // gradle dependencyUpdates
-    id 'com.diffplug.spotless' version '5.9.0'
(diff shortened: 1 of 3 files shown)


totorean:
BEFORE:
Total build time | 53.271s
No cache

AFTER:
Total build time | 6.729s

Tasks whose outputs were requested from cache : 69
 Hit 69 (100%)
 Local 69 (100%)
 Miss 0

ST-DDT (approved):
Looks good to me, but I'm not that familiar with gradle.
What are the potential side effects of this change?

@yidongnan What do you think?

yidongnan:
LGTM 👍

PR #520, opened 2021-04-16
Java 16 + gradle 7.0 compatibility

popfalushi:
Hello. Tried to build examples on gradle 6.8.3 and java 16 and build process failed for 3 reasons:
0. Gradle 6.8.3 failed with Could not compile initialization script 'C:\Users\popfa\AppData\Local\Temp\ijmapper.gradle'. Updated gradle to 7.0.
1. Lombok plugin needed update to be compatible with java 16. Changed plugin to io.freefair.lombok because it is recommended by Gradle.
2. Spotless plugin needed additional arguments in gradle.properties. For more information look here: https://github.com/diffplug/spotless/issues/834 . Plugin version got updated as well just because at first I thought that it will resolve the problem, but afterwards found issue on github. Didn't reverse plugin update after that.

After all changes local examples build and run ok, overall project also, javadoc is generated with getters/setters methods (checked on GrpcCodecDefinition class javadoc).


popfalushi:
As far as I understand, java 11 will pass the CI test and java 8 will be eternally failed because java 8 does not support --add-opens. Don't really know what to do with it. If java8 is essential than I suppose PR can be closed. Maybe in the future spotless plugin will do its job without need for add-opens, but at the same time lombok most likely will require it.

ST-DDT:
>     1. Changed plugin to io.freefair.lombok because it is recommended by Gradle.

Do you have a link for that? Or at least a comparison between these two?

ST-DDT:
> As far as I understand, java 11 will pass the CI test and java 8 will be eternally failed because java 8 does not support --add-opens. Don't really know what to do with it. If java8 is essential than I suppose PR can be closed. Maybe in the future spotless plugin will do its job without need for add-opens, but at the same time lombok most likely will require it.

Thats a though one. I'll have a look this weekend.

popfalushi:
> > ```
> > 1. Changed plugin to io.freefair.lombok because it is recommended by Gradle.
> > ```
> 
> Do you have a link for that? Or at least a comparison between these two?

My mistake - it is lombok project that recommends this plugin - gradle itself does not recommend anything. This plugin is referenced on official lombok page: https://projectlombok.org/setup/gradle and it is also under active development - it has milestone and stable version from April, the former plugin didn't get an update from last May.

FYI: Lombok right now is in the tough spot: https://github.com/rzwitserloot/lombok/issues/2681 . There were many discussions about it on reddit and hackernews.

ST-DDT:
IMO making our project compilable in the latest version is a good thing.
However, we should ensure that we generate Java 8 compatible jars (ClassVersion + Java Api), as there might still be developers out there that use that version for compatibility reasons.
Could you adjust the gradle file to always generate J8 compatible classes?

@yidongnan What do you think?

ST-DDT:
I included the gradle update in another PR that only updates dependencies #521, so that we can focus or changes here entirely to the J16 changes.

ST-DDT (changes requested):
Could you please update your PR for the latest changes?

popfalushi:
in a week or two - right now I'm on holiday. But what about compatibility with java 8? My PR was incompatible with java 8 and it will stay this way in the future. If compatibity with java 8 is a must than what's the point?

If you are ready to ditch java 8 - I'll rebase this branch on current state, no problem.

ST-DDT:
> If you are ready to ditch java 8 - I'll rebase this branch on current state, no problem.

I'm not sure about that yet. Enjoy your holidays, after that there is still plenty of time to decide on this.

ST-DDT:
Fixes #604

ST-DDT (approved):
Sorry for the long wait. Lets get this merged.

gradle.properties
```diff
@@ -5,9 +5,3 @@ org.gradle.caching=true
 org.gradle.vfs.watch=true
 org.gradle.daemon=true
 
-# for spotless java 16 compatibility: https://github.com/diffplug/spotless/issues/834
-org.gradle.jvmargs=--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED \
-  --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED \
-  --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED \
-  --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED \
-  --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
```
yidongnan: I think it's possible to add these jvm parameters depending on the Java version? I'll try it later

yidongnan: @ST-DDT What do you think if the Java version is greater than 15 and we disable the spotless plugin?

ST-DDT: I'd rather omit the java 17 compile time support for now and start compiling the project with JDK 11 (ClasssVersion still on Java 8) after the next release. Using the `javac --release 8` flag ensures that the produced releases are Java 8 compatible.
Then we can add the above settings to the gradle.properties and really support J17.

(I would like to release 2.13 this month, if both of our time permits.)

yidongnan: @ST-DDT ok

Issue #501, opened 2021-03-03
Doesn't allow to implement Grpc Core framework health.proto for health check


nehalshah50:
Following code errors with 
`
Caused by: java.lang.IllegalStateException: Found duplicate service implementation: grpc.health.v1.Health
	at net.devh.boot.grpc.server.serverfactory.AbstractGrpcServerFactory.configureServices(AbstractGrpcServerFactory.java:127)
`

`
@GrpcService
class HealthCheck : HealthGrpc.HealthImplBase(){

    override fun check(request: HealthCheckRequest, responseObserver: StreamObserver<HealthCheckResponse>) {
        coreLoggingService.Debug("Health Check ...")
        responseObserver.onNext(HealthCheckResponse.newBuilder().setStatus(HealthCheckResponse.ServingStatus.SERVING).build())
        responseObserver.onCompleted()
    }

    override fun watch(request: HealthCheckRequest, responseObserver: StreamObserver<HealthCheckResponse>) {
        coreLoggingService.Debug("Health Check - watch ...")
        responseObserver.onNext(HealthCheckResponse.newBuilder().setStatus(HealthCheckResponse.ServingStatus.SERVING).build())
        responseObserver.onCompleted()
    }
}

`

ST-DDT:
This library currently assumes, that you reuse the existing health service implementation (HealthServiceManager).
It's removal is planned for the next release. See also #411

PR #528, opened 2021-04-25
Make grpc-services.jar optional + remove grpc health service setup

### LINKED ISSUE #411: java.lang.ClassNotFoundException: io.grpc.services.HealthStatusManager

ST-DDT:
Fixes #411 , #501

Remove the grpc-service setup from the server factory making the library optional.

Fixed potential sensitive data exposure vulnerability.

### DIFF
--- a/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcHealthServiceAutoConfiguration.java
+++ b/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcHealthServiceAutoConfiguration.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2016-2021 Michael Zhang <yidongnan@gmail.com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+ * documentation files (the "Software"), to deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE
+ * WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package net.devh.boot.grpc.server.autoconfigure;
+
+import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import io.grpc.BindableService;
+import io.grpc.protobuf.services.ProtoReflectionService;
+import io.grpc.services.HealthStatusManager;
+import net.devh.boot.grpc.server.service.GrpcService;
+
+/**
+ * Auto configuration that sets up the grpc health service.
+ *
+ * @author Daniel Theuke (daniel.theuke@heuboe.de)
+ */
+@Configuration
+@ConditionalOnClass(ProtoReflectionService.class)
+@ConditionalOnProperty(prefix = "grpc.server", name = "health-service-enabled", matchIfMissing = true)
+@AutoConfigureBefore(GrpcServerFactoryAutoConfiguration.class)
+public class GrpcHealthServiceAutoConfiguration {
+
+    /**
+     * Creates a new HealthStatusManager instance.
+     *
+     * @return The newly created bean.
+     */
+    @Bean
+    @ConditionalOnMissingBean
+    HealthStatusManager healthStatusManager() {
+        return new HealthStatusManager();
+    }
+
+    @Bean
+    @GrpcService
+    BindableService grpcHealthService(final HealthStatusManager healthStatusManager) {
+        return healthStatusManager.getHealthService();
+    }
+
+}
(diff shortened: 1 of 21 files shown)


yidongnan:
@ST-DDT If the user is using `HealthCheckingLoadBalancer`, we directly delete the grpc health service, will it affect the grpc call?

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcReflectionServiceAutoConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+import io.grpc.protobuf.services.ProtoReflectionService;
+import net.devh.boot.grpc.server.service.GrpcService;
+
+/**
+ * Auto configuration that sets up the proto reflection service.
+ *
+ * @author Daniel Theuke (daniel.theuke@heuboe.de)
+ */
+@Configuration
+@ConditionalOnClass(ProtoReflectionService.class)
```
asarkar, line 36: Where is this class going to come from if not on classpath? I think the check should be a string, not class literal.

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/serverfactory/AbstractGrpcServerFactory.java
```diff
@@ -110,17 +99,6 @@ protected void configure(final T builder) {
     protected void configureServices(final T builder) {
         final Set<String> serviceNames = new LinkedHashSet<>();
 
-        // support health check
-        if (this.properties.isHealthServiceEnabled()) {
-            builder.addService(this.healthStatusManager.getHealthService());
-            serviceNames.add(HealthGrpc.getServiceDescriptor().getName());
```
asarkar, line 116: Now that the library no longer adds the health service, how do you recommend users cope with this change? Should they use a `GrpcServerConfigurer` to add it themselves?

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/config/GrpcServerProperties.java
```diff
@@ -180,14 +180,6 @@
...
     private DataSize maxInboundMetadataSize = null;
 
-    /**
-     * Whether gRPC health service is enabled or not. Defaults to {@code true}.
-     *
-     * @param healthServiceEnabled Whether gRPC health service is enabled.
-     * @return True, if the health service is enabled. False otherwise.
-     */
-    private boolean healthServiceEnabled = true;
-
```
asarkar, line 190: Instead of removing completely, can health not be enabled just like reflection?

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcReflectionServiceAutoConfiguration.java
ST-DDT: The annotation is analyzed by ASM before the class is loaded.
I tested it locally at it works as expected.

If its on a method level, then one has to use strings because the class is already loaded (or failed to be loaded) then.

grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/serverfactory/AbstractGrpcServerFactory.java
ST-DDT: I can add it back, but I wont populate the service with any defaults anymore, because blindly setting it to SERVING doesnt help with anything.

Just like any other service, you can add it as a simple bean:

````java
@Bean
BindableService healthService(HealthManager healthManager) {
    return healthManager.bindService();
}
````

I strongly recommend NOT adding it via ServerConfigurer as that bypasses all interceptors including security.
Which reminds me, I should add a hint/warning to them.

File: grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/config/GrpcServerProperties.java (hunk 3 above)
Comment (ST-DDT) [reply grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/config/GrpcServerProperties.java:190]: Sure, I will add it back.

**Comment by ST-DDT** [comment]
> @ST-DDT If the user is using `HealthCheckingLoadBalancer`, we directly delete the grpc health service, will it affect the grpc call?

I haven't looked into the implementation yet. I will check it.

**Review by yidongnan — APPROVED** [review]

File: grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcHealthServiceAutoConfiguration.java [hunk 4]
```diff
@@ -0,0 +1,60 @@
...
+import io.grpc.services.HealthStatusManager;
+import net.devh.boot.grpc.server.service.GrpcService;
+
+/**
+ * Auto configuration that sets up the grpc health service.
+ *
+ * @author Daniel Theuke (daniel.theuke@heuboe.de)
+ */
+@Configuration
+@ConditionalOnClass(ProtoReflectionService.class)
```
asarkar, line 38: @ST-DDT This condition seems wrong. AFAIK, health service has nothing to do with reflection; checking for `HealthStatusManager.class` makes more sense.

ST-DDT: Both classes are inside the same jar, so it shouldn't matter.
I will change it for clarity and in case of future refactors/splits.

PR #539, opened 2021-05-02
Condense/Simplify examples

ST-DDT:
- Merges all cloud discovery examples into a single example using profiles
  - Add example run with `consul`
- Remove bearer auth example, that is broken and was never fully functional


examples/cloud-grpc-server/src/main/resources/application-consul.yml
```diff
@@ -0,0 +1,7 @@
+#spring:
+#  cloud:
+#    consul:
+#      discovery:
+#        register: true
+#        hostname: localhost
+#        port: 8500
```
yidongnan, line 7: One small question, why are the configuration contents commented? Same problem in other yml file?

ST-DDT: Because they use the default value.
Might be confusing to have the commented out.
Should I un-comment them?

yidongnan: OK, it looks good now

yidongnan (approved):

PR #548, opened 2021-05-21
Support unix domain sockets

cfredri4:
This adds support for Unix Domain Sockets by specifying the address as: `unix:///my/socket`

(sorry for all the commits, I don't have a local build environment, please squash)


ST-DDT:
Some classes seems to be missing. Does these changes require a new (missing) dependency?

> io.netty.channel.unix.DomainSocketAddress
> io.netty.channel.epoll.EpollEventLoopGroup;

We should mention the new scheme/prefix in the server's properties' javadocs
and in the docs for both server and client.

Would you like to contribute that as well or should we take care of that?

cfredri4:
> Some classes seems to be missing. Does these changes require a new (missing) dependency?
> 
> > io.netty.channel.unix.DomainSocketAddress
> > io.netty.channel.epoll.EpollEventLoopGroup

Yes, it seems `netty-transport-native-epoll` is not included in non-shaded grpc-netty dependency.
What do you think is best? Have this feature with shaded-netty only? Or adding epoll as (optional) dependency, but which would then break if it is forgotten to be included?
I'm thinking the latter, and clearly documenting that epoll dependency is needed with non-shaded Netty (and that it only works on Linux, i.e. not on Windows or Mac). I'm not into Gradle (Maven ftw 😄) so please add this dependency if you agree.


> We should mention the new scheme/prefix in the server's properties' javadocs
> and in the docs for both server and client.
> 
> Would you like to contribute that as well or should we take care of that?

If you can link to where it should be documented then I can give it a shot (but I will not be offended if you take care of it; also my Chinese is rusty 😉).

ST-DDT:
> What do you think is best? Have this feature with shaded-netty only? Or adding epoll as (optional) dependency, but which would then break if it is forgotten to be included?

IMO netty and netty-shaded should support the same features.
Thus including it as optional dependency sounds like a valid choice.
We just have to ensure that this does not have an impact on the default setup where the dependency is missing.
Do you know a way to add automated tests that cover dependencies being present/absent?

> Maven ftw 😄

👍 

> I'm not into Gradle so please add this dependency if you agree.

I'll do that, if I have some free time. Hopefully next weekend.

> If you can link to where it should be documented then I can give it a shot (but I will not be offended if you take care of it; also my Chinese is rusty 😉).

I usually also omit that part...

The server part should probably be added here:
* https://github.com/yidongnan/grpc-spring-boot-starter/blob/ed6a0e50cb69caf2dfc42771a044a8fac3140904/docs/en/server/configuration.md#changing-the-server-port (after)
* https://github.com/yidongnan/grpc-spring-boot-starter/blob/ed6a0e50cb69caf2dfc42771a044a8fac3140904/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/config/GrpcServerProperties.java#L71

And the client part here:
* https://github.com/yidongnan/grpc-spring-boot-starter/blob/ed6a0e50cb69caf2dfc42771a044a8fac3140904/docs/en/client/configuration.md#choosing-the-target
* https://github.com/yidongnan/grpc-spring-boot-starter/blob/ed6a0e50cb69caf2dfc42771a044a8fac3140904/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/config/GrpcChannelProperties.java#L107

You can give it a try, otherwise I will add it along with the optional dependency. It might take a couple of extra days until I have some spare time though.

ST-DDT:
I found [this document](https://grpc.github.io/grpc/cpp/md_doc_naming.html) stating that both `unix:` and `unix://` are valid schemes. It also mentions `unix-abstract:`.
Do you know what the difference between them are and whether they are supported by grpc-java?

These seems to be some catches included in the code that I'll try to figure out.

ST-DDT:
@cfredri4 I polished this PR a bit. Could you please provide an answer the this question: Could https://github.com/yidongnan/grpc-spring-boot-starter/pull/548#issuecomment-853773220 ?

@yidongnan Do you consider this useful enough to be available by default?

ST-DDT (approved):

cfredri4:
> @cfredri4 I polished this PR a bit. Could you please provide an answer the this question: Could [#548 (comment)](https://github.com/yidongnan/grpc-spring-boot-starter/pull/548#issuecomment-853773220) ?

Interesting, I had not seen this. [Looking through the code](https://github.com/grpc/grpc-java/search?q=unix) it seems that grpc-java references unix sockets during tests and benchmarks, but there is no resolver to automatically parse "unix:" or equivalent into the correct SocketAddress.
There also seems to be some ambiguity with the prefix; the [benchmarks use "unix://" prefix](https://github.com/grpc/grpc-java/blob/50012832deeadf7edb84236f2174226ea9f38e9c/benchmarks/src/main/java/io/grpc/benchmarks/Utils.java#L61), while [xds uses "unix:"](https://github.com/grpc/grpc-java/blob/50012832deeadf7edb84236f2174226ea9f38e9c/xds/src/main/java/io/grpc/xds/internal/sds/SdsClient.java#L112) prefix. The latter is probably more correct.

yidongnan (approved):
LGTM

PR #578, opened 2021-09-02
Add logging response messages and other data to ClientLogInterceptor example

a-simeshin:
### Context 

Unfortunately on github and in the official documentation there are unfairly few examples on working with intercepting messages, and even less on intercepting data when responding from the server on the client side: response message, headers, status, trailers and other metadata.

I believe it would be a good option to show this work with a simple logging example.

### DIFF
--- a/examples/cloud-grpc-client/src/main/java/net/devh/boot/grpc/examples/cloud/client/LogGrpcInterceptor.java
+++ b/examples/cloud-grpc-client/src/main/java/net/devh/boot/grpc/examples/cloud/client/LogGrpcInterceptor.java
@@ -17,18 +17,22 @@
 
 package net.devh.boot.grpc.examples.cloud.client;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import io.grpc.CallOptions;
 import io.grpc.Channel;
 import io.grpc.ClientCall;
 import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall;
+import io.grpc.ForwardingClientCallListener;
+import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import net.devh.boot.grpc.client.interceptor.GrpcGlobalClientInterceptor;
 
 /**
- * Example {@link ClientInterceptor} that logs all called methods. In this example it is added to Spring's application
+ * Example {@link ClientInterceptor} that logs all called methods in INFO log level, also request and response messages,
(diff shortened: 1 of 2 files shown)


examples/cloud-grpc-client/src/main/java/net/devh/boot/grpc/examples/cloud/client/LogGrpcInterceptor.java
```diff
@@ -44,7 +49,39 @@
             final Channel next) {
 
         log.info(method.getFullMethodName());
-        return next.newCall(method, callOptions);
+        return new ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(next.newCall(method, callOptions)) {
+
+            @Override
+            public void sendMessage(ReqT message) {
+                log.debug("Request message: \n{}", message.toString());
```
ST-DDT, line 56: Please remove all the `.toString()` calls, as they waste performance when loggin is turned of and reduce readability.

ST-DDT (commented):
This is a very nice addition to the examples.
Could you please update the original log messages to include some additional text such as `Received call to {}`? That way they match the new log statements more and follow best practices.

examples/cloud-grpc-client/src/main/java/net/devh/boot/grpc/examples/cloud/client/LogGrpcInterceptor.java
a-simeshin: TO BE (for my own)
- [ ] log messages should follow pattern ``Received {action} with {action object}`` for example 
``Received call to sendMessage with message {}``
- [ ] remove all .toString()

Is that pattern correct? Sorry, I'm not a very experienced developer :(

ST-DDT: There is no real log pattern to follow as long as it done consistingly.
The most important part is that each log statement provides some kind of context for the log message.
Unfortunately, the original example only logged the called method without context. Your new log statements are good as they are, you only have upgrade the existing/old ones.

The `.toString()` is called conditionally by the logging implementation, sometimes with some additional formatting for lists/classes etc.

> Sorry, I'm not a very experienced developer :(

No problem at all. Everybody has to start somewhere. Reviewing and improving existing code is a very good way to improve your skills. (And it feels very nice to receive contributions once in a while 😉 )

a-simeshin: Thanks a lot for the explanation! I really appreciate it. 
Ready for a second review

ST-DDT (approved):
Looks good to me.

PR #579, opened 2021-09-02
Fix in docs client configuration examples with server interceptors

a-simeshin:
### Context

It seems to me that the client configuration page in the part of interceptors should contain examples with client, not server examples. Also, there is a mention of the ``GlobalClientInterceptorConfigurer`` in the documentation, but no example.

### What is done

- [x] Example with LogServerInterceptor changed to LogClientInterceptor with correct annotations in the client documentation
- [x] Added example with GlobalClientInterceptorConfigurer to the client documentation

### DIFF
--- a/docs/en/client/configuration.md
+++ b/docs/en/client/configuration.md
@@ -158,34 +158,48 @@ The following examples demonstrate how to use annotations to create a global cli
 
 ````java
 @Configuration
-public class ThirdPartyInterceptorConfig {}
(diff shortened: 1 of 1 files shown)


ST-DDT (approved):
Thanks for reporting and fixing this issue.
I put quite some effort into the docs, so it's good to know that there are people out there who read the docs careful enough to spot these errors.

PR #580, opened 2021-09-05
Extra annotation for gRPC clients/stubs integration with spring context

### LINKED ISSUE #463: Allow @GrpcClient on parameters of @Bean methods
**The problem**

We have a lib which is not tied to spring, exposing a class which needs a stub to be passed as a parameter. What I'd like is to use all the goodness from this lib when using the both of them. Right now this is more or less how I do it:
```java
@Bean
public NonSpringClient configNonSpringClient(@Value("${custom.path.to.grpc.address}") String address, @Value("${custom.path.to.grpc.port}") int port) {
    ManagedChannel channel = ManagedChannelBuilder.forAddress(address, port)
        .usePlaintext()
        .intercept(/* list of interceptors repeated from StubTransformer */)
        .build();
    MyGrpcServiceBlockingStub stub = MyGrpcService.newBlockingStub(channel);
    return new NonSpringClient(stub);
}
```
But I loose all the benefits from using `@GrpcClient`.

All in all, this is not blocking, but it would be really cool and more flexible if we could use `@GrpcClient` at the same places you can put a `@Value`, for example.

**The solution**

I want to be able to write the following:
```java
@Bean
public NonSpringClient configNonSpringClient(@GrpcClient("MY_STUB") MyGrpcServiceBlockingStub stub) {
    return new NonSpringClient(stub);
}
```
**Alternatives considered**

I'm not sure if what I'm asking is even possible, one alternative would be to give programmatic access to the stubs through an injectable class, something like:
```java
@Bean
public NonSpringClient configNonSpringClient(GrpcStubInjector injector) {
    return new NonSpringClient(injector.getStub("MY_STUB"));
}
```

a-simeshin:
### Context
Closes #463

There is a rather inconvenient problem described in https://github.com/yidongnan/grpc-spring-boot-starter/issues/463
``@GrpcClient`` is not a ``@Bean``, so we are encouraged to wrap it in a standard bean creation for quality of life.

```java
@Configuration
public class MyCustomGrpcBeanConfigh {

    @GrpcClient("client")
    private MyStub myStub;

    @Bean
    MyExternalBean myExternalBean() {
        return new MyExternalBean(myStub);
    }

}

@Configuration
public class MyAnotherCustomGrpcBeanConfig {

    @Bean
    MyAnotherExternalBean myAnotherExternalBean(@Autiwired MyExternalBean myExternalBean) {
        return new MyAnotherExternalBean (myExternalBean.getStub());
    }

}
```

### TO BE

```java
@Configuration
public class MyCustomGrpcBeanConfigh {

    @GrpcClient("client")
    private MyStub myStub;

}

@Configuration
public class MyAnotherCustomGrpcBeanConfig {

    @Bean
    MyAnotherExternalBean myAnotherExternalBean(@Autiwired MyStub myStub) {
        return new MyAnotherExternalBean (myStub);
    }

}
```

### TO DO
- [x] register client as bean after injection section into BeanPostProcessor
- [x] invoke autowire manually via AutowireCapableBeanFactory
- [x] make sure the bean is created in context, ``@Autowired`` and ``@Qualifier`` are available for method and field injection
- [x] proof it with unit test

### DIFF
--- a/docs/en/client/configuration.md
+++ b/docs/en/client/configuration.md
@@ -10,7 +10,6 @@ This section describes how you can configure your grpc-spring-boot-starter clien
   - [Choosing the Target](#choosing-the-target)
 - [Configuration via Beans](#configuration-via-beans)
   - [GrpcClientBean](#grpcclientbean)
-  - [GrpcClientBeans](#grpcclientbeans)
   - [GrpcChannelConfigurer](#grpcchannelconfigurer)
   - [ClientInterceptor](#clientinterceptor)
   - [StubFactory](#stubfactory)
--- a/docs/en/client/configuration.md
+++ b/docs/en/client/configuration.md
@@ -127,79 +127,37 @@ configuration class, but it's still possible.
 ````java
 @Configuration
 @GrpcClientBean(
-    clazz = TestServiceGrpc.TestServiceBlockingStub.class,
+    clazz = TestServiceBlockingStub.class,
     beanName = "blockingStub",
     client = @GrpcClient("test")
 )
-public static class YourCustomConfiguration {
+@GrpcClientBean(
+    clazz = FactoryMethodAccessibleStub.class,
+    beanName = "accessibleStub",
+    client = @GrpcClient("test"))
+public class YourCustomConfiguration {
 
     @Bean
-    FoobarService foobarService(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub) {
+    FooService fooServiceBean(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub) {
         return new FoobarService(blockingStub);
     }
+
 }
 
 @Service
 @AllArgsConsturtor
-public class FoobarService {
-    private TestServiceBlockingStub blockingStub;
+public class BarService {
 
-    public String receiveGreeting(String name) {
-        HelloRequest request = HelloRequest.newBuilder()
-                .setName(name)
-                .build();
-        return blockingStub.sayHello(request).getMessage();
-    }
-}
-````
-
-### GrpcClientBeans
-
-`@GrpcClientBeans` designed for registration multiple stubs to the spring context from single configuration class.
-
-````java
-@Configuration
-@GrpcClientBeans(value = {
-        @GrpcClientBean(
-                clazz = TestServiceGrpc.TestServiceBlockingStub.class,
-                beanName = "blockingStub",
-                client = @GrpcClient("test")),
-        @GrpcClientBean(
-                clazz = TestServiceGrpc.FactoryMethodAccessibleStub.class,
-                beanName = "accessibleStub",
-                client = @GrpcClient("test"))
-})
-public static class YourCustomConfiguration {
-
-    @Bean
-    FoobarService foobarService(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub,
-                                @Autowired TestServiceGrpc.FactoryMethodAccessibleStub accessibleStub) {
-        return new FoobarService(blockingStub, accessibleStub);
-    }
-}
-
-@Service
-@AllArgsConsturtor
-public class FoobarService {
-
-    private TestServiceBlockingStub blockingStub;
     private FactoryMethodAccessibleStub accessibleStub;
 
     public String receiveGreeting(String name) {
         HelloRequest request = HelloRequest.newBuilder()
                 .setName(name)
                 .build();
-        return blockingStub.sayHello(request).getMessage();
+        return accessibleStub.sayHello(request).getMessage();
     }
 
-    public String receiveAnotherGreeting(String name) {
-        HelloRequest request = HelloRequest.newBuilder()
-                .setName(name)
-                .build();
-        return accessibleStub.sayHi(request).getMessage();
-    }
 }
-````
 
 ### GrpcChannelConfigurer
 
--- a/docs/en/client/configuration.md
+++ b/docs/en/client/configuration.md
@@ -9,6 +9,8 @@ This section describes how you can configure your grpc-spring-boot-starter clien
 - [Configuration via Properties](#configuration-via-properties)
   - [Choosing the Target](#choosing-the-target)
 - [Configuration via Beans](#configuration-via-beans)
+  - [GrpcClientBean](#grpcclientbean)
+  - [GrpcClientBeans](#grpcclientbeans)
   - [GrpcChannelConfigurer](#grpcchannelconfigurer)
   - [ClientInterceptor](#clientinterceptor)
   - [StubFactory](#stubfactory)
@@ -114,6 +116,93 @@ extension points that exist in this library.
 First of all most of the beans can be replaced by custom ones, that you can configure in every way you want.
 If you don't wish to go that far, you can use classes such as `GrpcChannelConfigurer` and `StubTransformer` to configure
 the channels, stubs and other components without losing the features provided by this library.
+
+### GrpcClientBean
+
+This should significantly help with ``@Autowire`` and ``@Qualifier`` because default annotation``@GrpcClient`` is
+not designed for usage with spring 'injection' annotations.
+``@GrpcClientBean`` require annotation ``@Cofiguration`` or inherited, for example ``@TestConfiguraiton``.
+It is definitely not recommended using ``@GrpcClientBean`` and field annotated with ``@GrpcClient`` for same
+configuration class, but it`s still possible.
+
+````java
+@Configuration
+@GrpcClientBean(
+    clazz = TestServiceGrpc.TestServiceBlockingStub.class,
+    beanName = "blockingStub",
+    client = @GrpcClient("test")
+)
+public static class YourCustomConfiguration {
+
+    @Bean
+    FoobarService(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub) {
+        return new FoobarService(blockingStub);
+    }
+}
+
+@Service
+@AllArgsConsturtor
+public class FoobarService {
+    private TestServiceBlockingStub blockingStub;
+
+    public String receiveGreeting(String name) {
+        HelloRequest request = HelloRequest.newBuilder()
+                .setName(name)
+                .build();
+        return blockingStub.sayHello(request).getMessage();
+    }
+
+}
+````
+
+### GrpcClientBeans
+
+``@GrpcClientBeans`` designed for registration multiple stubs to the spring context from single configuration class.
+
+````java
+    @Configuration
+    @GrpcClientBeans(value = {
+            @GrpcClientBean(
+                    clazz = TestServiceGrpc.TestServiceBlockingStub.class,
+                    beanName = "blockingStub",
+                    client = @GrpcClient("test")),
+            @GrpcClientBean(
+                    clazz = TestServiceGrpc.FactoryMethodAccessibleStub.class,
+                    beanName = "accessibleStub",
+                    client = @GrpcClient("test"))
+    })
+    public static class YourCustomConfiguration {
+
+    @Bean
+    FoobarService(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub,
+                  @Autowired TestServiceGrpc.FactoryMethodAccessibleStub accessibleStub) {
+        return new FoobarService(blockingStub, accessibleStub);
+    }
+}
+
+@Service
+@AllArgsConsturtor
+public class FoobarService {
+
+    private TestServiceBlockingStub blockingStub;
+    private FactoryMethodAccessibleStub accessibleStub;
+
+    public String receiveGreeting(String name) {
+        HelloRequest request = HelloRequest.newBuilder()
+                .setName(name)
+                .build();
+        return blockingStub.sayHello(request).getMessage();
+    }
+
+    public String receiveAnotherGreeting(String name) {
+        HelloRequest request = HelloRequest.newBuilder()
+                .setName(name)
+                .build();
+        return accessibleStub.sayHi(request).getMessage();
+    }
+
+}
+````
 
 ### GrpcChannelConfigurer
 
--- a/docs/en/client/configuration.md
+++ b/docs/en/client/configuration.md
@@ -139,7 +139,7 @@ public class YourCustomConfiguration {
 
     @Bean
     FooService fooServiceBean(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub) {
-        return new FoobarService(blockingStub);
+        return new FooService(blockingStub);
     }
 
 }
@@ -158,6 +158,7 @@ public class BarService {
     }
 
 }
+````
 
 ### GrpcChannelConfigurer
 
--- a/docs/en/client/configuration.md
+++ b/docs/en/client/configuration.md
@@ -119,11 +119,10 @@ the channels, stubs and other components without losing the features provided by
 
 ### GrpcClientBean
 
-This should significantly help with ``@Autowire`` and ``@Qualifier`` because default annotation``@GrpcClient`` is
-not designed for usage with spring 'injection' annotations.
-``@GrpcClientBean`` require annotation ``@Cofiguration`` or inherited, for example ``@TestConfiguraiton``.
-It is definitely not recommended using ``@GrpcClientBean`` and field annotated with ``@GrpcClient`` for same
-configuration class, but it`s still possible.
+This should significantly help with `@Autowire` and `@Qualifier` because default annotation `@GrpcClient` is
+not designed for usage with spring 'injection' annotations. `@GrpcClientBean` require annotation `@Cofiguration`.
+It is definitely not recommended using `@GrpcClientBean` and field annotated with `@GrpcClient` for same
+configuration class, but it's still possible.
 
 ````java
 @Configuration
@@ -135,7 +134,7 @@ configuration class, but it`s still possible.
 public static class YourCustomConfiguration {
 
     @Bean
-    FoobarService(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub) {
+    FoobarService foobarService(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub) {
         return new FoobarService(blockingStub);
     }
 }
@@ -151,31 +150,30 @@ public class FoobarService {
                 .build();
         return blockingStub.sayHello(request).getMessage();
     }
-
 }
 ````
 
 ### GrpcClientBeans
 
-``@GrpcClientBeans`` designed for registration multiple stubs to the spring context from single configuration class.
+`@GrpcClientBeans` designed for registration multiple stubs to the spring context from single configuration class.
 
 ````java
-    @Configuration
-    @GrpcClientBeans(value = {
-            @GrpcClientBean(
-                    clazz = TestServiceGrpc.TestServiceBlockingStub.class,
-                    beanName = "blockingStub",
-                    client = @GrpcClient("test")),
-            @GrpcClientBean(
-                    clazz = TestServiceGrpc.FactoryMethodAccessibleStub.class,
-                    beanName = "accessibleStub",
-                    client = @GrpcClient("test"))
-    })
-    public static class YourCustomConfiguration {
+@Configuration
+@GrpcClientBeans(value = {
+        @GrpcClientBean(
+                clazz = TestServiceGrpc.TestServiceBlockingStub.class,
+                beanName = "blockingStub",
+                client = @GrpcClient("test")),
+        @GrpcClientBean(
+                clazz = TestServiceGrpc.FactoryMethodAccessibleStub.class,
+                beanName = "accessibleStub",
+                client = @GrpcClient("test"))
+})
+public static class YourCustomConfiguration {
 
     @Bean
-    FoobarService(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub,
-                  @Autowired TestServiceGrpc.FactoryMethodAccessibleStub accessibleStub) {
+    FoobarService foobarService(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub,
+                                @Autowired TestServiceGrpc.FactoryMethodAccessibleStub accessibleStub) {
         return new FoobarService(blockingStub, accessibleStub);
     }
 }
@@ -200,7 +198,6 @@ public class FoobarService {
                 .build();
         return accessibleStub.sayHi(request).getMessage();
     }
-
 }
 ````
 
--- a/docs/en/client/configuration.md
+++ b/docs/en/client/configuration.md
@@ -118,10 +118,11 @@ the channels, stubs and other components without losing the features provided by
 
 ### GrpcClientBean
 
-This should significantly help with `@Autowire` and `@Qualifier` because default annotation `@GrpcClient` is
-not designed for usage with spring 'injection' annotations. `@GrpcClientBean` require annotation `@Cofiguration`.
-It is definitely not recommended using `@GrpcClientBean` and field annotated with `@GrpcClient` for same
-configuration class, but it's still possible.
+This annotation is used to create injectable beans from your otherwise non-injectable `@GrpcClient` instances.
+The annotation can be repeatedly added to any of your `@Configuration` classes.
+
+> **Note:** We recommend using either `@GrpcClientBean`s or fields annotated with `@GrpcClient` throughout your
+> application, as mixing the two might cause confusion for future developers.
 
 ````java
 @Configuration
--- a/docs/en/client/getting-started.md
+++ b/docs/en/client/getting-started.md
@@ -107,7 +107,7 @@ If you don't wish to use any advanced features, then the first element is probab
   Supports `Channel`s, and all kinds of `Stub`s.
   Do not use `@GrpcClient` in conjunction with `@Autowired` or `@Inject`.
   Currently, it isn't supported for constructor and `@Bean` method parameters. For this case look below
-  to the``@GrpcClientBean``. \
+  to the `@GrpcClientBean`. \
   **Note:** Services provided by the same application can only be accessed/called in/after the
   `ApplicationStartedEvent`. Stubs connecting to services outside of the application can be used earlier; starting with
   `@PostConstruct` / `InitializingBean#afterPropertiesSet()`.
--- a/docs/en/client/getting-started.md
+++ b/docs/en/client/getting-started.md
@@ -181,8 +181,8 @@ public class FoobarService {
 Also you can feel free to inject stub with `@GrpcClientBean` with `@Configuration` for more wide usage in
 another services.
 
-> **Note:** It is definitely not recommended using ``@GrpcClientBean`` and field annotated with ``@GrpcClient`` like in
-> the previous example for same configuration class, but it`s still possible.
+> **Note:** We recommend using either `@GrpcClientBean`s or fields annotated with `@GrpcClient` throughout your
+> application, as mixing the two can cause confusion for future developers.
 
 ````java
 @Configuration
--- a/docs/en/client/getting-started.md
+++ b/docs/en/client/getting-started.md
@@ -106,10 +106,16 @@ If you don't wish to use any advanced features, then the first element is probab
   The annotation that marks fields and setters for auto injection of clients.
   Supports `Channel`s, and all kinds of `Stub`s.
   Do not use `@GrpcClient` in conjunction with `@Autowired` or `@Inject`.
-  Currently it isn't supported for constructor and `@Bean` method parameters. \
+  Currently, it isn't supported for constructor and `@Bean` method parameters. For this case look below
+  to the``@GrpcClientBean``. \
   **Note:** Services provided by the same application can only be accessed/called in/after the
   `ApplicationStartedEvent`. Stubs connecting to services outside of the application can be used earlier; starting with
   `@PostConstruct` / `InitializingBean#afterPropertiesSet()`.
+- [`@GrpcClientBean`](https://javadoc.io/page/net.devh/grpc-client-spring-boot-autoconfigure/latest/net/devh/boot/grpc/client/inject/GrpcClientBean.html):
+  The annotation should help with registration ``@GrpcClient`` to the spring context for usage with ``@Autowire`` and
+  ``@Qualifier``, required annotation ``@Configuration`` or inherited, for example ``@TestConfiguration``.
+- [`@GrpcClientBeans`](https://javadoc.io/page/net.devh/grpc-client-spring-boot-autoconfigure/latest/net/devh/boot/grpc/client/inject/GrpcClientBeans.html):
+  The annotation should help with multiple registration ``@GrpcClient`` to the spring context.
 - [`Channel`](https://javadoc.io/page/io.grpc/grpc-all/latest/io/grpc/Channel.html):
   The Channel is a connection pool for a single address. The target servers might serve multiple grpc-services though.
   The address will be resolved using a `NameResolver` and might point to a fixed or dynamic number of servers.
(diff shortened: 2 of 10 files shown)


ST-DDT:
I really like the idea of your change, but IMO the `@GrpcClient` bean in its current state supports multiple specific variations for a single client (name). Creating a bean for every `@GrpcClient` annotation would break existing code (duplicate bean definition) or behave randomly (e.g. due to the ranom bean initialization order).
How about adding a new annotation for that.

````java
/**
 * Annotation that can be added to `@Configuration` classes to create GrpcClient beans in the ApplicationContext.
 */
@Target(CLASS)
@Repeatable
public @interface GrpcClientBean {

    /** The type of the bean to create. */
    Class<?> clazz();

    /** The name of the bean to create. If empty, a name will be generated automatically based on the bean class and the client name. */
    String beanName() default "";

    /** The client definition used to create the bean. */
    GrpcClient definition();

}
````

Usage:

````java
@GrpcClientBean(clazz = SimpleServiceStub.class, definition = @GrpcClient("foobar"))
public class GrpcBeanConfig {

    MyAnotherExternalBean myExternalBean(SimpleServiceStub foobarSimpleServiceStub) {
        return new MyAnotherExternalBean (foobarSimpleServiceStub);
    }

}
````

I'm not sure whether we should add a `scope` option as the stubs themselves are immutable?

**Comment by ST-DDT** [comment]
See also https://github.com/yidongnan/grpc-spring-boot-starter/issues/463
( I plan to refactor the bean post processor bean a bit to use an `GrpcStubInjector` to split the stub generation from the stub injection, this would hardly have any impact on the implementation of this feature )

**Comment by a-simeshin** [comment]
> I really like the idea of your change, but IMO the `@GrpcClient` bean in its current state supports multiple specific variations for a single client (name). 

I thought about this problem. It seems to me that a good solution would be to declare the bean name as ``@GrpcClient value`` + ``class name`` of the stub, because it is impossible by means of this library to create two identical types of stubs for one client name.

Unfortunately, this idea has failed. 
I still don't understand why, when registering a bean name as ``name + injectionType.getSimpleName()``, then it cannot be found in the context by ``@Qualifier``, but when registering a bean name as ``Introspector.decapitalize(injectionType.getSimpleName())``, everything goes well. 
@ST-DDT, сould you be so kind to tell me why this is happening? It just blows my mind :(

> Creating a bean for every `@GrpcClient` annotation would break existing code (duplicate bean definition) or behave randomly (e.g. due to the ranom bean initialization order).

That is why "soft" registration of the bean is proposed with try catch and only warn. Without this, indeed, even in the examples of unit tests, it was clear that about 50% would break without proper softness. It seems to me that this solution should not break anything in existing projects, and tests confirm this, while allowing you to use all the advantages of ``@Bean`` in new projects.

> How about adding a new annotation for that.

To be honest, I don't like this solution, because for me all the beauty of this project is in 1 simple annotation that will do everything for you, then we will have more and more. On the other hand, we will be sure that nothing will break.
Anyway, @ST-DDT, I will trust your opinion and wait for your comment below. Will do it as you say

**Comment by ST-DDT** [comment]
> > I really like the idea of your change, but IMO the `@GrpcClient` bean in its current state supports multiple specific variations for a single client (name).
> 
> I thought about this problem. It seems to me that a good solution would be to declare the bean name as `@GrpcClient value` + `class name` of the stub, because it is impossible by means of this library to create two identical types of stubs for one client name.
> 
> Unfortunately, this idea has failed.
> I still don't understand why, when registering a bean name as `name + injectionType.getSimpleName()`, then it cannot be found in the context by `@Qualifier`, but when registering a bean name as `Introspector.decapitalize(injectionType.getSimpleName())`, everything goes well.
> @ST-DDT, сould you be so kind to tell me why this is happening? It just blows my mind :(

After some debugging I managed to get it to work. In my case I either forgot to update one of the two `@Qualifier` annotations or had typos in them (other character case etc).

I modified the code a bit to make it easier to debug:

````java

         final String beanName = annotation.value()+ injectionType.getSimpleName();
//        final String beanName = Introspector.decapitalize(injectionType.getSimpleName());
        try {
            final ConfigurableListableBeanFactory beanFactory =
                    ((ConfigurableApplicationContext) this.applicationContext).getBeanFactory();
            final Set<String> names = Sets.newHashSet(beanFactory.getBeanNamesIterator());
            beanFactory.registerSingleton(beanName, value);
            final Set<String> names2 = Sets.newHashSet(beanFactory.getBeanNamesIterator());
            names2.removeAll(names);
            log.warn("New Names: {}", names2);

            this.applicationContext.getAutowireCapableBeanFactory().autowireBean(value);
        } catch (final Exception e) {
            log.warn("Could not register and autowire bean: {}", beanName, e);
        }
````

IMO distinguishing them by client name is a requirement, especially since the following is potentially frequently used (especially if spread over multiple classes/libraries):

````java
        @GrpcClient("test")
        TestServiceGrpc.TestServiceBlockingStub blockingStub;
        
        @GrpcClient("fallback")
        TestServiceGrpc.TestServiceBlockingStub fallbackBlockingStub;
````

> > Creating a bean for every `@GrpcClient` annotation would break existing code (duplicate bean definition) or behave randomly (e.g. due to the ranom bean initialization order).
> 
> That is why "soft" registration of the bean is proposed with try catch and only warn. Without this, indeed, even in the examples of unit tests, it was clear that about 50% would break without proper softness. It seems to me that this solution should not break anything in existing projects, and tests confirm this, while allowing you to use all the advantages of `@Bean` in new projects.

You are right. Your soft registration does not break any existing code, but it writes plenty of warning to the logs.
(That is if you have used `@GrpcClient SomeStub` in more than one class.)
If I just run the tests, I get 265 of these warnings. Sure the test code does not reflect real code, but this huge number still indicates, that there is an issue there (84 with client disambiguation).

> > How about adding a new annotation for that.
> 
> To be honest, I don't like this solution, because for me all the beauty of this project is in 1 simple annotation that will do everything for you, then we will have more and more. On the other hand, we will be sure that nothing will break.

Thanks for sharing your opinion. I summarized the benefits of each variant below, please comment, if I forgot something.

**Pro OneForAll:**

````java
@GrpcClient("client")
private MyStub myStub;
````

* Single annotation for all purposes
* Easy to define

**Pro ExtraAnnotation:**

````java
@GrpcClientBean(clazz = MyStub.class, definition = @GrpcClient("client"))
````

* Unambiguous
* Allows declaring multiple beans of a single type
* Allows customization of bean names
* No need to define a field just to create a bean
* Easy to define

From this summary, I still prefer the ExtraAnnotation variant, but that's just my opinion.
@yidongnan , @Shinigami92 what do you think would be the best way here? (feel free to propose other solutions as well)

**Review by ST-DDT — CHANGES_REQUESTED** [review]
I consulted with @Shinigami92 and we came to the conclusion, that the unambiguity of the bean definition is more important than having only a single annotation for all.

Please change your PR to create an extra annotation for grpc clients/stubs that will be added to the application context.

**Comment by a-simeshin** [comment]
> I consulted with @Shinigami92 and we came to the conclusion, that the unambiguity of the bean definition is more important than having only a single annotation for all.
> 
> Please change your PR to create an extra annotation for grpc clients/stubs that will be added to the application context.

Thank you very much for the research after reanalysis, the solution with additional annotation looks better for me now. I`ll start ASAP on this week

Before starting, I propose to have your agreement with solution, that we should add annotation, which add to the context a bean
```java
/**
 * Annotation that can be added to `@Configuration` classes to create GrpcClient beans in the ApplicationContext.
 */
@Target(CLASS)
@Repeatable
public @interface GrpcClientBean {

    /** The type of the bean to create. */
    Class<?> clazz();

    /** The name of the bean to create. If empty, a name will be generated automatically based on the bean class and the client name. */
    String beanName() default "";

    /** The client definition used to create the bean. */
    GrpcClient definition();

}
```

But there are several questions that are not entirely clear to me:
1.  If the bean name is not specified, then should I create a bean with name ``GrpcClient.value()`` + ``clazz.getSimpleName()`` ?
2.  Should I call manually ``autowireBean(beanValue)`` after bin creation?
3.  Do I need to create a separate BeanPostProcessor class that will work only with this annotation, separately from ``@GrpcClient`` ?
4. Should I check, that class with annotation ``@GrpcBeanClient`` also annotated with ``@Configuration``? Or we can offer a softer solution?
4.  How a situation should be handled when an annotated field with ``@GrpcClient`` already declared manually in the class with annotation ``@GrpcBeanClient`` ?

ST-DDT:
> 2. Should I call manually `autowireBean(beanValue)` after bin creation?

I'm not sure about that. Will the beans get autowired if not called?
IMO it is not that important for now, as the `@grpcClient` injections aren't autowired till now.

> 3. Do I need to create a separate BeanPostProcessor class that will work only with this annotation, separately from `@GrpcClient` ?

No, IMO the existing post processor is good as it. It will get trimmed a bit when the injector will be split into a separate class.

> 4. Should I check, that class with annotation `@GrpcBeanClient` also annotated with `@Configuration`? Or we can offer a softer solution?

That's a good question. My first thought was we should check for the `@GrpcBeanClient` annotation, and if it isn't an annotated with `@Configuration`, then we should throw an exception. While it is probably not necessary to add the restriction, I hope to establish some good coding practices with that.

> 5. How a situation should be handled when an annotated field with `@GrpcClient` already declared manually in the class with annotation `@GrpcBeanClient` ?

IMO we should not add any additional limitations on these, so it will be possible to to have both `@GrpcClientBean` on the class itself and `@GrpcClient` on its fields.

Any suggestions regarding field names? I'm really not sure which names would be best.

- `clazz` or `type` or `beanType`
- `beanName` or `name`
- `definition` or `client`

a-simeshin:
> Any suggestions regarding field names? I'm really not sure which names would be best.
> 
> * `clazz` or `type` or `beanType`
> * `beanName` or `name`
> * `definition` or `client`

I think that simple and concise names just awesome. So ``clazz`` and ``client`` would be nice, but ``beanName`` allows to avoid misunderstandings with the name in the annotation ``@GrpcClient``, because it`s very likely that someone has already created a bean with the client's name in real projects (even me).

a-simeshin:
@ST-DDT would you be so kind to code review again? ``GrpcClientBeanInjectionTest`` with example of usage, also you were definitely right, living without beans hassle and warns in the log is much more better.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -223,8 +244,13 @@ private GrpcChannelFactory getChannelFactory() {
...
             return injectionType.cast(stub);
         } else {
-            throw new InvalidPropertyException(injectionTarget.getDeclaringClass(), injectionTarget.getName(),
-                    "Unsupported type " + injectionType.getName());
+            if (injectionTarget != null) {
+                throw new InvalidPropertyException(injectionTarget.getDeclaringClass(), injectionTarget.getName(),
+                        "Unsupported type " + injectionType.getName());
+            } else {
+                throw new InvalidPropertyException(injectionType.getDeclaringClass(), injectionType.getName(),
+                        "Unsupported type " + injectionType.getName());
```
ST-DDT, line 252: Have you checked, whether the exception message contains a reference to the annotated `@Configuration`?

ST-DDT (changes requested):
Looks good to me, I have just some small suggestions regarding the code readability.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -97,6 +101,22 @@ public Object postProcessBeforeInitialization(final Object bean, final String be
                             processInjectionPoint(method, paramTypes[0], annotation));
                 }
             }
+
+            for (final GrpcClientBean beanClientIterator : clazz.getAnnotationsByType(GrpcClientBean.class)) {
+                final String beanNameToCreate = beanClientIterator.beanName().isEmpty()
+                        ? beanClientIterator.client().value() + beanClientIterator.clazz().getSimpleName()
+                        : beanClientIterator.beanName();
```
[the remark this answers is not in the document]
a-simeshin: Done

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -97,6 +101,22 @@ public Object postProcessBeforeInitialization(final Object bean, final String be
...
+                try {
+                    final ConfigurableListableBeanFactory beanFactory =
+                            ((ConfigurableApplicationContext) this.applicationContext).getBeanFactory();
```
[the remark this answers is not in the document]
a-simeshin: Done

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -97,6 +101,22 @@ public Object postProcessBeforeInitialization(final Object bean, final String be
...
+                    final Object beanValue =
+                            processInjectionPoint(null, beanClientIterator.clazz(), beanClientIterator.client());
+                    beanFactory.registerSingleton(beanNameToCreate, beanValue);
+                } catch (final Exception e) {
+                    log.warn("Could not register and autowire bean: {}", beanNameToCreate, e);
```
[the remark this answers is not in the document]
a-simeshin: Done, but unfortunately I have no idea how to catch exceptions through ``AssertThrows`` at the stage of forming the context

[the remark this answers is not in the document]
ST-DDT: No problem.

These work like plain JUnit tests without spring (no spring boot test annotations on the test class).
We start spring with the required configuration just inside the test.

It will look similar to this:

````java
@Test
void testFail() {
    SpringApplication app = new SpringApplication(TestConfig.class);
    assertThrows(SomeException.class, () -> app.run());
}
````

Do you wish to give it a try?

File: grpc-client-spring-boot-autoconfigure/build.gradle [hunk 5]
```diff
@@ -15,6 +15,7 @@ dependencies {
 
     api project(':grpc-common-spring-boot')
     api 'org.springframework.boot:spring-boot-starter'
+    api 'org.springframework.boot:spring-boot-starter-test'
```
ST-DDT, line 18: We should not depend on this test dependency at runtime.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
a-simeshin: I added check for annotations ``@Configuration`` or ``@TestConfiguration``.
In addition added the name of the class where the error has occurred to the exception text.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -318,4 +324,16 @@ private String getBeanName(final GrpcClientBean grpcClientBean) {
...
 
+    /**
+     * The method is used to check for the presence of an annotation {@link Configuration} or {@link TestConfiguration}
+     *
+     * @param clazz instance of the class
+     * @return does the class have an annotation or not
+     */
+    private boolean isAnnotatedWithConfiguration(final Class<?> clazz) {
+        final Configuration[] configurationAnnotation = clazz.getAnnotationsByType(Configuration.class);
+        final TestConfiguration[] testConfigurationAnnotation = clazz.getAnnotationsByType(TestConfiguration.class);
```
ST-DDT, line 335: You can avoid this check by searching for the `@Configuration` annotation being meta present.

````java
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@Documented
@Configuration // <---
@TestComponent
public @interface TestConfiguration {
````

See also `AnnotationUtils.findAnnotation(...)`

**Review by ST-DDT — CHANGES_REQUESTED** [review]

File: grpc-client-spring-boot-autoconfigure/build.gradle (hunk 5 above)
Comment (a-simeshin) [reply grpc-client-spring-boot-autoconfigure/build.gradle]: Removed

File: grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java (hunk 6 above)
Comment (a-simeshin) [reply grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java]: Thanks for advice! ``AnnotationUtils.findAnnotation(...)`` works well, I didn't even know that it can work at the class level.
Done

File: grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java (hunk 4 above)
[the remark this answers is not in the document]
Comment (a-simeshin) [reply grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java]: It works С:
Add test to check bean conflict and test for simultaneous use ``@GrpcClient`` with custom bean and ``@GrpcClientBean`` in one ``@Configuration`` class.

**Comment by a-simeshin** [comment]
@ST-DDT would you be so kind to final review?

File: docs/en/client/configuration.md [hunk 7]
```diff
@@ -115,6 +117,93 @@ First of all most of the beans can be replaced by custom ones, that you can conf
 If you don't wish to go that far, you can use classes such as `GrpcChannelConfigurer` and `StubTransformer` to configure
 the channels, stubs and other components without losing the features provided by this library.
 
+### GrpcClientBean
+
+This should significantly help with ``@Autowire`` and ``@Qualifier`` because default annotation``@GrpcClient`` is
+not designed for usage with spring 'injection' annotations.
+``@GrpcClientBean`` require annotation ``@Cofiguration`` or inherited, for example ``@TestConfiguraiton``.
+It is definitely not recommended using ``@GrpcClientBean`` and field annotated with ``@GrpcClient`` for same
+configuration class, but it`s still possible.
```
ST-DDT, line 126:
```suggestion
configuration class, but it's still possible.
```

docs/en/client/configuration.md
```diff
@@ -115,6 +117,93 @@ First of all most of the beans can be replaced by custom ones, that you can conf
 If you don't wish to go that far, you can use classes such as `GrpcChannelConfigurer` and `StubTransformer` to configure
 the channels, stubs and other components without losing the features provided by this library.
 
+### GrpcClientBean
+
+This should significantly help with ``@Autowire`` and ``@Qualifier`` because default annotation``@GrpcClient`` is
```
ST-DDT, line 122: AFAIK these can all use single backticks '`'

docs/en/client/configuration.md
```diff
@@ -115,6 +117,93 @@ First of all most of the beans can be replaced by custom ones, that you can conf
...
+                    client = @GrpcClient("test"))
+    })
+    public static class YourCustomConfiguration {
+
+    @Bean
+    FoobarService(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub,
+                  @Autowired TestServiceGrpc.FactoryMethodAccessibleStub accessibleStub) {
+        return new FoobarService(blockingStub, accessibleStub);
+    }
+}
```
ST-DDT, line 181: Something is wrong with the indentation here.

docs/en/client/getting-started.md
```diff
@@ -106,10 +106,16 @@ If you don't wish to use any advanced features, then the first element is probab
...
   Do not use `@GrpcClient` in conjunction with `@Autowired` or `@Inject`.
-  Currently it isn't supported for constructor and `@Bean` method parameters. \
+  Currently, it isn't supported for constructor and `@Bean` method parameters. For this case look below
+  to the``@GrpcClientBean``. \
   **Note:** Services provided by the same application can only be accessed/called in/after the
   `ApplicationStartedEvent`. Stubs connecting to services outside of the application can be used earlier; starting with
   `@PostConstruct` / `InitializingBean#afterPropertiesSet()`.
+- [`@GrpcClientBean`](https://javadoc.io/page/net.devh/grpc-client-spring-boot-autoconfigure/latest/net/devh/boot/grpc/client/inject/GrpcClientBean.html):
+  The annotation should help with registration ``@GrpcClient`` to the spring context for usage with ``@Autowire`` and
+  ``@Qualifier``, required annotation ``@Configuration`` or inherited, for example ``@TestConfiguration``.
```
ST-DDT, line 116: IMO there is no need to mention inherited `@Configuration`s (e.g. `@TestConfiguration`), as I assume that it matches the expectation. (Both here and above)

docs/en/client/getting-started.md
```diff
@@ -106,10 +106,16 @@ If you don't wish to use any advanced features, then the first element is probab
...
+- [`@GrpcClientBeans`](https://javadoc.io/page/net.devh/grpc-client-spring-boot-autoconfigure/latest/net/devh/boot/grpc/client/inject/GrpcClientBeans.html):
+  The annotation should help with multiple registration ``@GrpcClient`` to the spring context.
```
ST-DDT, line 118: This section is unneccessary, just mention that the `@GrpcClientBean` is repeatable.

docs/en/client/getting-started.md
```diff
@@ -167,13 +173,48 @@ public class FoobarService {
...
     }
 
 }
 ````
 
+Also you can feel free to inject stub with ``@GrpcClientBean`` with ``@Configuration`` for more wide usage in
+another services.
+
+> **Note:** It is definitely not recommended using ``@GrpcClientBean`` and field annotated with ``@GrpcClient`` like in
+> the first example for same configuration class, but it`s still possible.
```
ST-DDT, line 187:
```suggestion
> the previous example for same configuration class, but it`s still possible.
```

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -53,6 +57,7 @@
  * @author Michael (yidongnan@gmail.com)
  * @author Daniel Theuke (daniel.theuke@heuboe.de)
  */
+@Slf4j
```
ST-DDT, line 60: AFAICT this logger is currently unused

ST-DDT (changes requested):
Sorry, for the delay with the review. I'm currently quite busy so I can only do a fast pass. I hope that I have more time next weekend.

docs/en/client/configuration.md
a-simeshin: Done

docs/en/client/configuration.md
a-simeshin: Replaced

docs/en/client/configuration.md
a-simeshin: Fixed

docs/en/client/getting-started.md
a-simeshin: Removed

docs/en/client/getting-started.md
a-simeshin: Done

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
a-simeshin: Removed annotation and import

docs/en/client/getting-started.md
a-simeshin: Removed in both files

docs/en/client/configuration.md
```diff
@@ -115,6 +117,90 @@ First of all most of the beans can be replaced by custom ones, that you can conf
...
+    }
+
+    public String receiveAnotherGreeting(String name) {
+        HelloRequest request = HelloRequest.newBuilder()
+                .setName(name)
+                .build();
+        return accessibleStub.sayHi(request).getMessage();
+    }
+}
+````
```
ST-DDT, line 202: I would recommend merging both chapters into one, as the `@GrpcClientBeans` is not that important for the developer.

```suggestion
````java
@Configuration
@GrpcClientBean(
    clazz = TestServiceBlockingStub.class,
    beanName = "blockingStub",
    client = @GrpcClient("test")
)
@GrpcClientBean(
    clazz = FactoryMethodAccessibleStub.class,
    beanName = "accessibleStub",
    client = @GrpcClient("test"))
public class YourCustomConfiguration {

    @Bean
    FooService fooServiceBean(@Autowired TestServiceGrpc.TestServiceBlockingStub blockingStub) {
        return new FoobarService(blockingStub);
    }

}

@Service
@AllArgsConsturtor
public class BarService {

    private FactoryMethodAccessibleStub accessibleStub;

    public String receiveGreeting(String name) {
        HelloRequest request = HelloRequest.newBuilder()
                .setName(name)
                .build();
        return accessibleStub.sayHello(request).getMessage();
    }

}
````
```

docs/en/client/configuration.md
```diff
@@ -9,6 +9,8 @@ This section describes how you can configure your grpc-spring-boot-starter clien
 - [Configuration via Properties](#configuration-via-properties)
   - [Choosing the Target](#choosing-the-target)
 - [Configuration via Beans](#configuration-via-beans)
+  - [GrpcClientBean](#grpcclientbean)
+  - [GrpcClientBeans](#grpcclientbeans)
```
ST-DDT, line 13:
```suggestion
```

docs/en/client/configuration.md
```diff
@@ -115,6 +117,90 @@ First of all most of the beans can be replaced by custom ones, that you can conf
 If you don't wish to go that far, you can use classes such as `GrpcChannelConfigurer` and `StubTransformer` to configure
 the channels, stubs and other components without losing the features provided by this library.
 
+### GrpcClientBean
+
+This should significantly help with `@Autowire` and `@Qualifier` because default annotation `@GrpcClient` is
+not designed for usage with spring 'injection' annotations. `@GrpcClientBean` require annotation `@Cofiguration`.
+It is definitely not recommended using `@GrpcClientBean` and field annotated with `@GrpcClient` for same
+configuration class, but it's still possible.
```
ST-DDT, line 125:
```suggestion
This annotation is used to create injectable beans from your otherwise non-injectable `@GrpcClient` instances.
The annotation can be repeatedly added to any of your `@Configuration` classes.

> **Note:** We recommend using either `@GrpcClientBean`s or fields annotated with `@GrpcClient` throughout your
> application, as mixing the two might cause confusion for future developers.
```

docs/en/client/getting-started.md
```diff
@@ -106,10 +106,14 @@ If you don't wish to use any advanced features, then the first element is probab
...
   Do not use `@GrpcClient` in conjunction with `@Autowired` or `@Inject`.
-  Currently it isn't supported for constructor and `@Bean` method parameters. \
+  Currently, it isn't supported for constructor and `@Bean` method parameters. For this case look below
+  to the``@GrpcClientBean``. \
   **Note:** Services provided by the same application can only be accessed/called in/after the
   `ApplicationStartedEvent`. Stubs connecting to services outside of the application can be used earlier; starting with
   `@PostConstruct` / `InitializingBean#afterPropertiesSet()`.
+- [`@GrpcClientBean`](https://javadoc.io/page/net.devh/grpc-client-spring-boot-autoconfigure/latest/net/devh/boot/grpc/client/inject/GrpcClientBean.html):
+  The annotation should help with registration `@GrpcClient` to the spring context for usage with `@Autowire` and
+  `@Qualifier`, required annotation `@Configuration`. Also, this annotation is repeatable using `@GrpcClientBeans`.
```
ST-DDT, line 116:
```suggestion
  The annotation helps to register `@GrpcClient` beans in the Spring context to be used with `@Autowired` and
  `@Qualifier`. The annotation can be repeatedly added to any of your `@Configuration` classes.
```

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -97,6 +103,25 @@ public Object postProcessBeforeInitialization(final Object bean, final String be
                             processInjectionPoint(method, paramTypes[0], annotation));
                 }
             }
+
+            if (isAnnotatedWithConfiguration(clazz)) {
+                for (final GrpcClientBean beanClientIterator : clazz.getAnnotationsByType(GrpcClientBean.class)) {
```
ST-DDT, line 108: How about calling the field simply `annotation`, as it isn't really an `Iterator`?

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -265,4 +296,41 @@ private GrpcChannelFactory getChannelFactory() {
...
+            this.configurableBeanFactory = ((ConfigurableApplicationContext) this.applicationContext).getBeanFactory();
+        }
+        return this.configurableBeanFactory;
+    }
+
+    /**
+     * The method is used to determine with which name the bean should be registered in the context
+     *
+     * @param grpcClientBean annotation value
+     * @return name
```
ST-DDT, line 315:
```suggestion
     * Gets the bean name from the given annotation.
     *
     * @param grpcClientBean The annotation to extract it from.
     * @return The extracted name.
```

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -265,4 +296,41 @@ private GrpcChannelFactory getChannelFactory() {
...
+        } else {
+            return grpcClientBean.client().value() + grpcClientBean.clazz().getSimpleName();
+        }
+    }
+
+    /**
+     * The method is used to check for the presence of an annotation {@link Configuration}
+     *
+     * @param clazz instance of the class
+     * @return does the class have an annotation or not
```
ST-DDT, line 329:
```suggestion
     * Checks whether the given class is annotated with {@link Configuration}.
     *
     * @param clazz The class to check.
     * @return True, if the given class is annotated with {@link Configuration}. False otherwise.
```

ST-DDT (changes requested):
Final review before merge. 👍

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
a-simeshin: You are definitely right! Renamed

a-simeshin:
@ST-DDT it seems that we have a flaky test - MetricCollectingInterceptorTest > testMetricsFailedCall() failed in different builds with just md changes:

- https://github.com/yidongnan/grpc-spring-boot-starter/pull/580/checks?check_run_id=3635062828
- https://github.com/yidongnan/grpc-spring-boot-starter/pull/580/checks?check_run_id=3635074122
- https://github.com/yidongnan/grpc-spring-boot-starter/pull/580/checks?check_run_id=3635118686

Have I to commit something to rerun build or there is more another "handy" way?
Should I create issue and try to fix and different PR?

ST-DDT:
> @ST-DDT it seems that we have a flaky test - MetricCollectingInterceptorTest > testMetricsFailedCall() failed in different builds with just md changes:

I am aware of that test, I already improved it multiple times to "fail less", but the error still prevails and I'm not sure why. Currently I get a ~1% chance of a failure in my local builds. Sometimes I suspect that it only happens at specific times.

> Have I to commit something to rerun build or there is more another "handy" way?

I have a re-run button in travis, that I can use to execute the checks again. Not sure, whether you have it as well.

> Should I create issue and try to fix and different PR?

Feel free to have a try (once this one is merged).

ST-DDT (approved):

ST-DDT:
Thanks for your contribution!

dileep-alla-by:
i am not able to initialize grpc stub by using @bean annotation 

error:- Parameter 0 of constructor in com.grpc.ContractRatesClientService required a bean of type 'com.grpc.v1.ReactorContractRatesGrpc$ReactorContractRatesStub' that could not be found. 

can  you help whats is the issue?

@GrpcClientBean(clazz = ReactorContractRatesStub.class, client = @GrpcClient("rating-data-service"))
@Configuration
public class RatingOrchestrationConfig {

    @Bean
    public ContractRatesClientService contractRatesClient(ReactorContractRatesStub contractRatesStub, RatesProtobufMapper ratesProtobufMapper) {
        return new ContractRatesClientService(contractRatesStub, ratesProtobufMapper);
    }
}


@Service
@Slf4j
@RequiredArgsConstructor
public class ContractRatesClientService {

    private final ReactorContractRatesGrpc.ReactorContractRatesStub contractRatesStub;
    private final RatesProtobufMapper ratesProtobufMapper;
}

ST-DDT:
Please try do move the `@GrpcClientBean` to its own config and order `RatingOrchestrationConfig` after that config.

It looks like I need to improve the initialization/bean declaration order somehow.

------------------

EDIT: Please try removing `@Service` from the service, as this might interfere with bean initialization order. (Or order it after the `RatingOrchestrationConfig`)

PR #642, opened 2022-02-28
Fix creation of Bearer Token using CallCredentialsHelper.bearerAuth(Supplier<String>)

vitorluis:
Hello fellas, first of all, thanks for this really nice starter for gRPC.

I was going through the documentation about how to secure the gRPC endpoints and there I followed the following code snippet to add the bearer token on clients:

```@Bean
CallCredentials bearerAuthForwardingCredentials() {
    return CallCredentialsHelper.bearerAuth(() -> KeycloakSecurityContext.getTokenString());
}
```

The problem is that inside the `CallCredentialsHelper.bearerAuth` instead of getting the token using the `get` method of the `Supplier` it was simply passing the `Supplier` itself (calling `toString()` implicitly). When I was debugging the server to check if the token was coming correctly and the `AuthenticationProvider` was doing its job I have stumbled upon this.

With that being said, I have committed the fix for this. Please, if possible push it to release version so we all have the fix.

Thanks. :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ buildscript {
         }
     }
     ext {
-        projectVersion = '2.14.0-SNAPSHOT'
+        projectVersion = '2.14.1-SNAPSHOT'
 
         // https://github.com/grpc/grpc-java/releases
         grpcVersion = '1.43.2'
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ buildscript {
         }
     }
     ext {
-        projectVersion = '2.14.1-SNAPSHOT'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
         }
     }
     ext {
-        projectVersion = '2.14.0-SNAPSHOT'
+        projectVersion = '2.14.1-SNAPSHOT'
```
ST-DDT, line 11:
```suggestion
        projectVersion = '2.14.0-SNAPSHOT'
```

ST-DDT (approved):

PR #645, opened 2022-03-21
Correct tag names according to Micrometer's docs

seckin206:
See. https://micrometer.io/docs/concepts#_tag_naming

With the current implementation and using official prometheus exporter,
we end up with label names `methodType` and `statusCode`. These labels
don't follow the snake_case naming convention that prometheus follows
for labels.
Using dot for word separator makes micrometer automatically convert
label names to snake case out of the box. 
See. https://github.com/micrometer-metrics/micrometer/blob/main/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheus/PrometheusNamingConvention.java

### DIFF
--- a/docs/en/actuator.md
+++ b/docs/en/actuator.md
@@ -59,7 +59,7 @@ Once the dependencies are added grpc-spring-boot-starter will automatically conf
 
 - `service`: The requested grpc service name (using protobuf name)
 - `method`: The requested grpc method name (using protobuf name)
-- `methodType`: The type of the requested grpc method.
+- `method.type`: The type of the requested grpc method.
 
 ### Timer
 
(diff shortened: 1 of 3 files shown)


rubda (approved):

ST-DDT:
This looks good to me.
Please also create a PR for micrometer and link it here:
https://github.com/micrometer-metrics/micrometer/blob/e4dfd439d20d36a778e688f2f21d9a84fafcbeaf/micrometer-binders/src/main/java/io/micrometer/binder/grpc/AbstractMetricCollectingInterceptor.java#L52-L59
As this implementation will be replaced with the upstream one mid term.

seckin206:
Opened the PR against micrometer: https://github.com/micrometer-metrics/micrometer/pull/3088

PR #658, opened 2022-04-27
Replace deprecated io.grpc.services.HealthStatusManager

marcindabrowski:
io.grpc.services.HealthStatusManager is deprecated and io.grpc.protobuf.services.HealthStatusManager should be used instead

### DIFF
--- a/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcHealthServiceAutoConfiguration.java
+++ b/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcHealthServiceAutoConfiguration.java
@@ -25,7 +25,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
 import io.grpc.BindableService;
-import io.grpc.services.HealthStatusManager;
+import io.grpc.protobuf.services.HealthStatusManager;
 import net.devh.boot.grpc.server.service.GrpcService;
 
 /**


marcindabrowski:
I don't know why it is failing, because locally on my macOS it is working. And I don't have permission to rerun it. I think that `MetricCollectingInterceptorTest.testMetricsCancelledCall` could be flaky.

Besides, I don't understand the idea of testing it using java 8 and 11 because in your gradle config Java 8 is set as target compatibility. And you are using toolchain and compatibility options at the same time. It doesn't make sense.
When you are using toolchains the java provided to run Gradle does not matter. It is used only to run Gradle, and to compile only when it is the same as the version set in the toolchain. So in the case of java 11 it will download Java 8 to compile the code.

And it is better to use https://github.com/gradle/gradle-build-action and it is good to use https://github.com/gradle/wrapper-validation-action. I can provide needed PR for that.

ST-DDT:
I reran it and now it passes. The `MetricCollectingInterceptorTest` have a chance of falling here on GitHub, but I'm unable to reproduce it and thus fix it locally.

> Besides, I don't understand the idea of testing it using java 8 and 11 because in your gradle config Java 8 is set as target compatibility.

This is mostly there to ensure that there are no library isssues with J11 (J9 plus removed some classes from the jre/jdk):
https://github.com/yidongnan/grpc-spring-boot-starter/blob/5e4e5fcfd96a80b157408a0f4c8f2108634d8787/grpc-client-spring-boot-autoconfigure/build.gradle#L23

I think it is fine to switch to J11 as the build jdk, but it should still produce J8 compatible jars.
@yidongnan What do you think?

> And it is better to use [gradle/gradle-build-action](https://github.com/gradle/gradle-build-action) and it is good to use [gradle/wrapper-validation-action](https://github.com/gradle/wrapper-validation-action). I can provide needed PR for that.

I would appreciate that, as I'm not familar with those.

ST-DDT (approved):

yidongnan:
> I think it is fine to switch to J11 as the build jdk, but it should still produce J8 compatible jars.
Or maybe even J17?
@yidongnan What do you think?

I think it's good to switch to j11 or j17 as build jdk, but we have to make sure the produce jar is compatible with j8

marcindabrowski:
Using the toolchain set to 8 will produce java 8 code. You can also add manifest entry with target JDK tag like this:

```
"X-Compile-Source-JDK"   : tasks.compileJava.sourceCompatibility,
"X-Compile-Target-JDK"   : tasks.compileJava.targetCompatibility,
```

PR #683, opened 2022-05-12
Fix kubernetes config doc

coltstrgj:
[This member variable](https://github.com/yidongnan/grpc-spring-boot-starter/blob/master/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/config/GrpcChannelsProperties.java#L48) is called "client" (not "clients") so spring will fail to populate the values as documented. 

Locally I tried:
```
grpc:
  client:
    my-server:
      address: dns:///my-server-using-client
  clients:
    my-server:
      address: dns:///my-server-using-client-with-s
```
And confirmed that the address is not populated with an "s" provided.

### DIFF
--- a/docs/en/kubernetes.md
+++ b/docs/en/kubernetes.md
@@ -51,15 +51,15 @@ spec:
 ````properties
 ## Choose your matching variant
 # Same namespace (target port=80 or derived by DNS-SVC)
-grpc.clients.my-grpc-server-app.address=dns:///my-grpc-server-app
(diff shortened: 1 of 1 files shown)


ST-DDT (approved):
Nicely spotted!

yidongnan (approved):

coltstrgj:
Thanks for getting to that so fast. Nice project by the way, it's been a huge help.

ST-DDT:
That is nice to hear ❤️ . Thanks again for your contribution. 👍

PR #736, opened 2022-09-03
Replace metric implementation with external/official one

### LINKED ISSUE #544: Replace metric implementation with micrometer native one
**The problem**

Micrometer provides a metric collection implementation. We should use that instead of reimplementing it ourselves.

**The solution**

Use the implementation from micrometer.
https://github.com/micrometer-metrics/micrometer/pull/2323
We already have an optional dependency to that library so nothing will really change here.

**Alternatives considered**

Retain our own implementation. This comes at a risk that it will eventually diverge from the micrometer one.

ST-DDT:
Fixes #544 

~~I need to fix the annotated `@Order` behavior first.~~ Done


ST-DDT:
Waiting on #737

yidongnan (approved):
LGTM

PR #737, opened 2022-09-03
Support `@Order` on interceptor bean factory methods

### LINKED ISSUE #627: Allow order Interceptors by using @Order annotation in the @Bean definition
**The problem**

We have a spring boot application that uses ServerInterceptors that we don't own. We found that the interceptors get called in a deterministic order based on the @Order annotation in the ServerInterceptor class. The problem I found is that if the @Order annotation is added to the @Bean definition it won't have any effect

**The solution**

I would like to be able to have a @Configuration class where I can define one @Bean per ServerInterceptor I want to register. I do want to be able to define the @Order as part of the bean definition so I could tell the library in which order I want them being called.

**Alternatives considered**

Another option is having the possibility to overwrite that order using GrpcServerConfigurer or GlobalServerInterceptorConfigurer

ST-DDT:
Fixes #627


yidongnan (approved):
LGTM

PR #749, opened 2022-09-19
Add support for client stub constructor injection

xJoeWoo:
This starter is lack of support for Constructor Injection of client stubs. CI is quite a common usage in Spring projects(especially projects with Kotlin) to make service class neat and easy to test. 

In old ways, multiple `@GrpcClientBean`s must be used on configuration class to register client stubs as Spring Bean. So that client stubs can be injected as a bean.

This PR implements a new way for simpler CI. [Example](https://github.com/xJoeWoo/grpc-spring-boot-starter/blob/constructor-inject/tests/src/test/java/net/devh/boot/grpc/test/inject/GrpcClientConstructorInjectionTest.java#L54)

The basic implement mechanism are:
1. `beanName` field was added to `@GrpcClient`, and make this annotation able to mark on constructor parameters;
2. Search all beans for all constructor parameters annotated with `@GrpcClient`, by `GrpcClientConstructorInjectBeanFactoryPostProcessor`. Collect grpc client config together with parameter type.
3. Build a client stub bean registrations class `GrpcClientConstructorInjection` then register it to bean factory;
4. Registering client stub beans in existing post processing class `GrpcClientBeanPostProcessor`, by [`GrpcClientBeanPostProcessor#initGrpClientConstructorInjects()`](https://github.com/xJoeWoo/grpc-spring-boot-starter/blob/constructor-inject/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java#L147).

Maybe **BREAKING CHANGES:**
1. Names of client stub beans must not the same if they are declared by `@GrpcClientBean`, even they have same class. This PR changes it to allow registering same name client stub bean **if** they have same class. [Source here](https://github.com/xJoeWoo/grpc-spring-boot-starter/blob/constructor-inject/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java#L179)

Also, Kotlin was introduced for Constructor Injection test.


grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
```diff
@@ -134,6 +141,17 @@ private void processMethods(final Class<?> clazz, final Object bean) {
...
+    /**
+     * Triggers registering grpc client beans from GrpcClientConstructorInjects in bean factory.
+     */
+    private void initGrpClientConstructorInjects() {
+        try {
+            for (GrpcClientConstructorInjection.GrpcClientBeanInjection injection : getConfigurableBeanFactory().getBean(GrpcClientConstructorInjection.class)) {
+                registerGrpcClientBean(injection, injection.getInjectClazz());
+            }
+        } catch (Exception ignored) {
+        }
```
ST-DDT, line 153: Why would we ignore exceptions here and which?

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientConstructorInjection.java
```diff
@@ -0,0 +1,59 @@
+package net.devh.boot.grpc.client.inject;
+
+import java.lang.annotation.Annotation;
+import java.util.ArrayList;
+import java.util.Iterator;
+
+class GrpcClientConstructorInjection implements Iterable<GrpcClientConstructorInjection.GrpcClientBeanInjection> {
```
ST-DDT, line 7: Why don't we use a list directly?

tests/src/test/java/net/devh/boot/grpc/test/inject/GrpcClientBeanInjectionNegativeTest.java
```diff
@@ -90,7 +90,7 @@ void mixedGrpcClientBeanAndFieldTest() {
             beanName = "duplicateStub",
             client = @GrpcClient("test"))
     @GrpcClientBean(
-            clazz = TestServiceGrpc.TestServiceBlockingStub.class,
+            clazz = TestServiceGrpc.TestServiceFutureStub.class,
```
ST-DDT, line 93: I have to think about the implications of this.
Maybe this should fail, because the interceptor config might be different on each bean.

tests/src/test/kotlin/net/devh/boot/grpc/test/kotlin/inject/KotlinGrpcClientInjectionTest.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    @GrpcClient("anotherTest", beanName = "anotherBlockingStub")
+    val anotherBlockingStub: TestServiceGrpc.TestServiceBlockingStub,
+
+    @GrpcClient("unnamed")
+    val unnamedTestServiceBlockingStub: TestServiceGrpc.TestServiceBlockingStub,
+)
+
+@Component
+class KotlinConstructorInjectAnotherBean(
```
ST-DDT, line 32: Could you rename these a bit and add some little javadoc what the difference between the two is?
Please also add a test for the old field injection.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientConstructorInjection.java
```diff
@@ -0,0 +1,59 @@
...
+            return client;
+        }
+
+        @Override
+        public Class<? extends Annotation> annotationType() {
+            return GrpcClientBean.class;
+        }
+
+        public Class<?> getInjectClazz() {
+            return injectClazz;
```
ST-DDT, line 45: Maybe call this targetClass or sourceClass?

ST-DDT (changes requested):
Awesome, this feature is on my wish-list since a long time.
I have to think about some subtile implications such as having two beans with the same type but different names. I'm not sure whether spring resolves them correctly.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientBeanPostProcessor.java
xJoeWoo: Mainly I want to ignore the `NoSuchBeanDefinitionException` from `getBean` then return. Yep, wrap the whole loop in try block is wrong.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientConstructorInjectBeanFactoryPostProcessor.java
```diff
@@ -0,0 +1,41 @@
...
+
+        // Use bean name to get bean class to avoid triggering bean init
+        beanFactory.getBeanNamesIterator().forEachRemaining(beanName -> {
+            Class<?> clazz = beanFactory.getType(beanName);
+            if (clazz == null) {
+                return;
+            }
+
+            // Search for GrpcClient annotation in all parameters of all constructors
+            for (Constructor<?> constructor : clazz.getDeclaredConstructors()) {
```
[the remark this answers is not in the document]
xJoeWoo: Good idea for looking for the constructor Spring would use. I'll check whether there's way to do that.

xJoeWoo:
I didn't evaluate the full impact to interceptor or other stub processing steps. Maybe some researches need to be made by maintainers.

ST-DDT:
Please run `gradle spotlessApply` and commit the changes.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientConstructorInjectBeanFactoryPostProcessor.java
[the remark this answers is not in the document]
ST-DDT: Untested suggestion: We could maybe get the bean definitions and then check whether [getConstructorArgumentValues()](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/config/BeanDefinition.html#getConstructorArgumentValues--)...[getSource()](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/config/ConstructorArgumentValues.ValueHolder.html#getSource--) or something along the way returns something useful.

There also exists a [AnnotatedGenericBeanDefinition](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/annotation/AnnotatedGenericBeanDefinition.html) subclass, maybe that grants access to the annotations.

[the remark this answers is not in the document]
xJoeWoo: I tried `getConstructorArgumentValues()` before, however, it always returns empty info that made this way not working.
 
image

To reduce performance loss from scanning all bean, I also tried to make an new annotation `@GrpcClientInject` on the classes requiring stubs injection, retrieved less bean classes by [`getBeanNamesForAnnotation()`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/support/DefaultListableBeanFactory.html#getBeanNamesForAnnotation-java.lang.Class-). It was a bit faster but the `@GrpcClientInject` annotation itself was easily missed, even during me writing my test codes. Instead of a bit performance improvement, reducing learning cost for users maybe more important.

ST-DDT:
Did you commit all changes from spotlessApply?

xJoeWoo:
> Did you commit all changes from spotlessApply?

Yep

ST-DDT:
The tests are failing.

xJoeWoo:
> The tests are failing.

I missed that, also I ran `test` task in root gradle project and every tests seemed ok now.

ST-DDT (commented):
I will have an in depth look at this and some extended testing this weekend.

There are a couple of things that I'm not sure about yet:
- Should we merge the PostProcessors? IMO they are very similar as they work on the same annotation and even exchange data. I'm not sure which implications that might have an the `initGrpClientConstructorInjects()` method though.
- I'm not sure whether/how spring determines that the GrpcClient Bean that is created for a constructor parameter is actually intended to be used for that parameter (if there are multiple parameters with the same class).
- whether there is a way to do this without adding a bean name to it
- whether there is a way to support bean factory methods/do this in a more generic way

This a very "technical" PR that challenges my knowledge about Spring's internals and thus it might take a few days to review.
Once again, thanks for this PR.

xJoeWoo:
It's my pleasure to help this project better 😄. I'm not quite an expert on Spring, maybe my ideas are not the perfect solution.

- The two PostProcessors perform their actions on different phases of launching spring. The bean factory postprocessor collects stubs requirements by constructors, **before** any beans are initialized. Otherwise due to missing bean corresponding to constructor parameter, the missing bean exception will be thrown when create a bean. The original processor runs **after** bean initialization, and reusing bean register codes from @GrpcClientBean. The initGrpClientConstructorInjects() method can be moved to first one processor but register codes will be repeated or another util class/reference to the other processor is needed.

- Multi same classes bean injection must specify which bean should be use. Usually we use @Qualifier(old way) or parameter name(new way, param name is same as bean name) to set a specific bean to parameter. Examples can be found in GrpcClient#beanName, creating a grpc stub bean with specific bean name, and make the name of constructor parameter same as bean name.

- It would be nice if spring provides some kind of definitions/configurers that can let us pass a specific instance to constructor parameter. However I'm lack of acknowledges about such tools 😂. So the existed client stub bean method is selected.

ST-DDT:
I had a look at this and I noticed two shortcomings that needs to be addressed.

- The bean name must match the constructor parameter name, and doing that can be confusing. To avoid accidental invalid bean assignments, the bean name should be derived from the parameter name and cannot be specified manually.
- The bean registration has to be unequivocal meaning that only if the bean (with the same name) was created by this library and it was created using the exact same configuration (annotation), then it is allowed. Otherwise a BeanCreationException should be thrown (or bean name conflict or whatever matches best...). It should be clearly documented in the `@GrpcClient` bean that if it is used in a constructor, then it will create a bean with the name of the parameter.

-----------------

I also considered explicitly setting the constructor argument values on the bean definition (there is a setter for the value), maybe this works and it isn't required to create a named bean anymore. I didn't have time to test it, but I would like to evaluate it before we merge this solution.

xJoeWoo:
I realized `ConstructorArgumentValues` in `BeanDefinition` is the tool I was looking for binding client stub to constructor parameter. `beanName` field on `@GrpcClient` had removed, and no beans will be created on latest commit.

ST-DDT:
Awesome, thisnlooks very promising. I will review this in the next few days.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientConstructorInjectionBeanFactoryPostProcessor.java
```diff
@@ -0,0 +1,68 @@
...
+                                    clazz,
+                                    beanFactory.getBeanDefinition(beanName),
+                                    i);
+
+                    grpcClientConstructorInjection.add(registry);
+                }
+            }
+        });
+
+        beanFactory.registerSingleton("grpcClientConstructorInjection", grpcClientConstructorInjection);
```
ST-DDT, line 66: Maybe skip the bean registration, if it is empty.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/inject/GrpcClientConstructorInjectionBeanFactoryPostProcessor.java
```diff
@@ -0,0 +1,68 @@
...
+
+        // Use bean name to get bean class to avoid triggering bean init
+        beanFactory.getBeanNamesIterator().forEachRemaining(beanName -> {
+            Class<?> clazz = beanFactory.getType(beanName);
+            if (clazz == null) {
+                return;
+            }
+
+            // Search for GrpcClient annotation in all parameters of all constructors
+            for (Constructor<?> constructor : clazz.getDeclaredConstructors()) {
```
ST-DDT, line 44: Is it possible to also analyze the bean factory methods here?
If `getFactoryBeanName()` and `getFactoryMethodName()` is set, then it should be preferred over the constructor analysis.
We should also limit ourselves to a single constructor to avoid issues, if there are multiple annotated parameters in different constructors.
I will search a bit to check how spring determines the constructor to use.

ST-DDT: Done

ST-DDT (approved):

yidongnan (approved):
LGTM

PR #812, opened 2022-12-22
Autoconfigure for updated Sleuth Brave class

clamey:
Between Sleuth 2.x and 3.x, the Brave Auto Configuration class package and name changed.  This updates the `GrpcCommonTraceAutoConfiguration` to wait to load after either version is present.

Currently, people who try to use Sleuth 3.x and grpc-spring-boot-starter might run into problems, like I did, if `GrpcCommonTraceAutoConfiguration` runs first.  It will create the `grpcTracing` bean, and then `BraveAutoConfiguration` will try to create a bean with the same name and fail.


ST-DDT (approved):

dsyer:
Probably duplicated in #829?

dsyer:
#829 was merged. Isn’t this redundant now?

ST-DDT:
Yes you are right.

Closing as obsolete

PR #829, opened 2023-01-20
Bump micrometer version to latest

### LINKED ISSUE #828: Upgrade Micrometer to latest release
The version of Micrometer currently building with this project is old and doesn't contain the new tracing/observation APIs. If we upgrade to 1.10.3 (or later) there shouldn't be any problems for the build or for existing users because existing APIs are unchanged, and it will clear a path to supporting Spring Boot 3 properly.

dsyer:
Fixes #828

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@ bin/
 .settings/
 .factorypath
 .checkstyle
+
+## VSCode
+.vscode/
+.devcontainer/
--- a/build.gradle
+++ b/build.gradle
@@ -186,9 +186,11 @@ allprojects { project ->
             }
         }
 
+        ext['micrometer.version'] = '1.10.3'
+
         ext {
-            micrometerVersion = '1.10.3'
             // not explicitly needed in subprojects, as the BOM for Spring Boot sets this version
+            micrometerVersion = dependencyManagement.importedProperties['micrometer.version']
             springFrameworkVersion = dependencyManagement.importedProperties['spring-framework.version']
             springSecurityVersion = dependencyManagement.importedProperties['spring-security.version']
             springCloudCommonsVersion = dependencyManagement.importedProperties['spring-cloud-commons.version']
--- a/build.gradle
+++ b/build.gradle
@@ -186,8 +186,6 @@ allprojects { project ->
             }
         }
 
-        ext['micrometer.version'] = '1.10.3'
-
         ext {
             // not explicitly needed in subprojects, as the BOM for Spring Boot sets this version
             micrometerVersion = dependencyManagement.importedProperties['micrometer.version']
--- a/build.gradle
+++ b/build.gradle
@@ -187,8 +187,8 @@ allprojects { project ->
         }
 
         ext {
-            micrometerVersion = dependencyManagement.importedProperties['micrometer.version']
-            // not explicitly needed in subprojects, as the BOM for Sprint Boot sets this version
+            micrometerVersion = '1.10.3'
+            // not explicitly needed in subprojects, as the BOM for Spring Boot sets this version
             springFrameworkVersion = dependencyManagement.importedProperties['spring-framework.version']
             springSecurityVersion = dependencyManagement.importedProperties['spring-security.version']
             springCloudCommonsVersion = dependencyManagement.importedProperties['spring-cloud-commons.version']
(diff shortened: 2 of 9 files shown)


dsyer:
It's a step in the direction of supporting Spring Boot 3. Whether this tiny change is a good idea or not really hinges on the approach the project maintainers want to take to that. If you want a smooth transition for users I think it should be possible to support Spring Boot 3 and 2.7 with a single release, and one of the steps required will be to upgrade Micrometer, while keeping support for older versions. This is the first baby step in that direction. If you prefer a different approach maybe you could outline it here or in one of the other issue discussions.

ST-DDT:
> and one of the steps required will be to upgrade Micrometer,

Thats the point I dont understand. Why is it required to update micrometer?

dsyer:
Because Spring Boot 3 uses micrometer features not available before 1.10, and because Spring Cloud Sleuth no longer exists and has effectively been replaced with Micrometer 1.10. Without this upgrade there’s no path for Spring Boot users to get tracing when they migrate to 3.0.

ST-DDT:
> Without this upgrade there’s no path for Spring Boot users to get tracing when they migrate to 3.0.

If they use the spring boot v3 parent bom, then they will automatically have the micrometer update. And if they dont, they can still update micrometer themselves. Depending on the order they are importing the dependencies in they might even get the micrometer update anyway.
Am I missing something?

dsyer:
Apparently. There’s no more sleuth so we need to add micrometer tracing for grpc to this library. There’s no way to do that without an upgrade, but it can be done in such a way that existing Spring Boot 2.7 apps won’t be affected (I believe).

ST-DDT:
> need to add micrometer tracing for grpc to this library

Since it is an optional dependency the user has to add it to their dependencies anyway.

dsyer:
It’s an optional dependency but this library contains a autoconfiguration for tracing, and users will expect that the same level of feature will be available for Spring Boot 3 as with Spring Cloud Sleuth. Again, this requires an upgrade of Micrometer.

ST-DDT:
So instead of just upgrading micrometer you want to add some bean config as well to automatically setup the tracing beans?

dsyer:
I think that’s the best way forward, but you’re the one making decisions. I’m not going to propose any configuration changes until we agree it’s a good idea. I didn’t include those changes partly because I wanted to sound you out first, and partly because there are other blockers for Spring Boot 3 migration (`SocketUtils` at least, possibly others).

dsyer:
I added a new commit with some autoconfiguration changes. I tested the server configuration in a sample. There's no tracing sample in this project and also no Spring Boot 3 sample, so I didn't want to add that too at the same time.

dsyer:
This is still marked as "waiting for feedback" but I'm not sure what is needed. Can we move this forward please?

ST-DDT:
> Can we move this forward please?

I'm currently busy and don't have time to work on this project. 😿

yidongnan:
I read the discussion above and it looks like this change will make it incompatible with spring boot 2.7, is it possible to make it compatible with both 2.7 and 3.0 while tracing is working?

dsyer:
I think you misunderstood. The whole point is to make Spring Boot 3 and 2.7 work (you still have to pick one and add the right dependencies).

yidongnan:
> I added a new commit with some autoconfiguration changes. I tested the server configuration in a sample. There's no tracing sample in this project and also no Spring Boot 3 sample, so I didn't want to add that too at the same time.

Sorry, I'm not very familiar with spring boot 3.0 at the moment and I see compatibility issues mentioned here.

I understand that micrometer was introduced to enable spring boot 3.0 applications to support grpc tracing, and if spring boot 2.7 is used and tracing still works (using spring cloud sleuth), then I think this PR is very valuable. Is there something I am missing?

dsyer:
Don't think so. That pretty much sums it up - with Spring Boot 2.7 you need Sleuth and with Spring Boot 3.0 it's just Micrometer.

yidongnan:
Would you be able to take some time to merge the master branch and update it?

dsyer:
I rebased and the tests are green. Haven't looked into the details it much.

yidongnan (approved):
LGTM, I have tested it, and it appears to be working as expected.

PR #877, opened 2023-04-17
Don't try to create grpcTracing bean if sleuth is not active

genuss:
Hello again!
I found this inconsistency when trying to use this starter with spring boot 3. With 3.0 version there are cases when `brave.Tracing` is on classpath, although no bean is created. That's default behaviour in tests for example. There may be (mostly theoretically) such cases with spring boot 2.x too.
I didn't create any tests, as this bug only shows up with 3.x branch.


ST-DDT (approved):

ST-DDT:
I don't have a real testing setup for this, so I have to just go by the code.

amitbansal26:
any updates on this ?

PR #957, opened 2023-09-04
Fix classNotFoundException when using old version of io.micrometer:mi…

echemezov:
Classes io.micrometer.core.instrument.binder.grpc.MetricCollectingServerInterceptor and io.micrometer.core.instrument.binder.grpc.MetricCollectingClientInterceptor are not available in old versions of  io.micrometer:micrometer-core (for example in 1.5.1) but io.micrometer.core.instrument.MeterRegistry used as a condition for GrpcClientMetricAutoConfiguration and GrpcServerMetricAutoConfiguration exists.

In result, application fails on start with ClassNotFoundException when there is MeterRegistry but no grpc interceptor classes.


yidongnan (approved):
LGTM

PR #970, opened 2023-09-28
Support springboot 3.x

yidongnan:
- [ ] Built with Java 17、21
- [ ] Support Spring Boot 3.x
- [ ] Support Spring Cloud 2022.0.x
- [ ] Support Spring Security


grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/SocketUtils.java
```diff
@@ -0,0 +1,320 @@
...
+ * port it is currently using.
+ *
+ * @author Sam Brannen
+ * @author Ben Hale
+ * @author Arjen Poutsma
+ * @author Gunnar Hillert
+ * @author Gary Russell
+ * @since 4.0
+ */
+public class SocketUtils {
```
genuss, line 50: As far as I could guess, this class was borrowed from spring framework 5.x. In case you're interested, with 6.x you can still use almost the same one without copying, it's called `org.springframework.test.util.TestSocketUtils`. It certainly breaks spring boot's 2.7 compatibility, but you do it anyway in other parts

yidongnan: I understand that there's the `org.springframework.test.util.TestSocketUtils` class, but if we were to use it directly, the project would need to have a dependency on `org.springframework:spring-test`. Since this pertains to just one file, I'm currently leaning towards copying that file for now.

yidongnan: Regarding the compatibility issue with Spring Boot 2.7, I believe we may need to consider eventually letting go of it and making a full transition to 3.x. I think version 3.0.0 should be a suitable choice for this transition.

genuss: Oh, right. The dependency on spring-test is definitely not desired for this case

PR #995, opened 2023-11-25
fix: run dynamic call credentials via executor

### LINKED ISSUE #958: Maybe resolve a supplier in the DynamicSecurityHeaderCallCredentials in an executor to avoid blocking the thread
CallCredentials documentation suggests using asynchronous way of setting headers in case there is a blocking function in the code (e.g. a network call).

I would change [this](https://github.com/yidongnan/grpc-spring-boot-starter/blob/9495d5e1f825ad384e96917bdb3a30b09373bd57/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/security/CallCredentialsHelper.java#L341-L345)

```java
        @Override
        public void applyRequestMetadata(final RequestInfo requestInfo, final Executor appExecutor,
                final MetadataApplier applier) {
            applier.apply(this.extraHeadersSupplier.get());
        }
```

To something like in [this example](https://github.com/grpc/grpc-java/blob/master/examples/example-jwt-auth/src/main/java/io/grpc/examples/jwtauth/JwtCredential.java#L39-L62)

```java
        @Override
        public void applyRequestMetadata(final RequestInfo requestInfo, final Executor appExecutor,
                final MetadataApplier applier) {
            executor.execute(new Runnable() {
                @Override
                public void run() {
                    try {
                        applier.apply(this.extraHeadersSupplier.get());
                    } catch (Throwable t) {
                        applier.fail(Status.UNAUTHENTICATED.withCause(e));
                    }
                }
            });
        }
```

ST-DDT:
Fixes #958

- #958

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/security/CallCredentialsHelper.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/security/CallCredentialsHelper.java
@@ -340,7 +340,13 @@ public class CallCredentialsHelper {
         @Override
         public void applyRequestMetadata(final RequestInfo requestInfo, final Executor appExecutor,
                 final MetadataApplier applier) {
-            applier.apply(this.extraHeadersSupplier.get());
+            appExecutor.execute(() -> {
+                try {
+                    applier.apply(this.extraHeadersSupplier.get());
+                } catch (RuntimeException e) {
+                    applier.fail(Status.UNAUTHENTICATED.withCause(e));
+                }
+            });
         }
 
         @Override


fengli79 (approved):

ST-DDT:
So would it be better to revert the change and just add a hint to the javadocs?

Or maybe introduce a new variant that it is explicitly called "async" that takes a `(executor) -> CompletableFuture<String>`?
That way users can return `CompletableFuture.completedFuture()` if nothing async needs to be done.

````java
private final Function<Executor, CompletableFuture<String>> extraHeadersFutureSupplier;

@Override
        public void applyRequestMetadata(final RequestInfo requestInfo, final Executor appExecutor,
                final MetadataApplier applier) {

            extraHeadersFutureSupplier.apply(executor).whenComplete((header, e) -> {
                if (e == null) {
                    final Metadata extraHeaders = new Metadata();
                    extraHeaders.put(AUTHORIZATION_HEADER, header);
                    applier.apply( extraHeaders);
                } else {
                    applier.fail(Status.UNAUTHENTICATED.withCause(e));
                }
            });
        }
````

On the other hand you could just let the user implement that temselves.

fengli79:
Yes, I feel the dynamic credential doesn't provide too much value to the end users. This is an important enough detail (cached/non-cached/async/blocking) which should be exposed to the implementer of the credential. Either way works, but I personally like the ideal to just let the user implement that themselves.

PR #1010, opened 2023-12-06
feat: Add gRPC Metrics

DNVindhya:


ST-DDT:
AFAICT these metrics are not specifically spring related, so they could be used by other bindings as well.
I think we could add it to the OpenTelemetry bindings repo directly that way other libraries would benefit from it as well and we don't risk having a different implementation from the rest.
We could still have the required autoconfiguration for the metrics in this library.

Or are there some already here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/grpc-1.6/library ?

fengli79:
@ST-DDT We are trying to implement the gRPC A66 metrics spec, with micrometer. We won't introduce any OTel dependency at this moment, and the instrumentation + exporting are fully implemented in micrometer. And later on, based on the micrometer instrumentation, OpenTelemetry may provide an end to end user experience in a non-invasive way, such as using a shim, etc.

I see you are the contributor of the gRPC metrics in micrometer as well, could you help to review and maybe accelerate the approval process there (we can implement in micrometer directly)?

ST-DDT:
> We are trying to implement the gRPC A66 metrics spec, with micrometer. We won't introduce any OTel dependency at this moment, and the instrumentation + exporting are fully implemented in micrometer.

Oh, I for some reason I have mistaken this for OT.

> I see you are the contributor of the gRPC metrics in micrometer as well, could you help to review and maybe accelerate the approval process there (we can implement in micrometer directly)?

Sure. Just ping me there and I'll have a look.

DNVindhya:
Closing this draft PR, as the changes have been merged in #1021.

PR #1031, opened 2024-01-17
feat: Add gRPC server metrics

DNVindhya:
This PR adds server metrics defined in gRFC [A66: OpenTelemetry Metrics](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md) to gRPC-Spring. Even though the gRFC mentions OpenTelmetry, to keep it in sync with existing metrics, this PR as well as future ones will use Micrometer for instrumentation.

Instrumentations to collect following server metrics are being introduced :
- `grpc.server.call.started`
- `grpc.server.call.sent_total_compressed_message_size`
- `grpc.server.call.rcvd_total_compressed_message_size`
- `grpc.server.call.duration`

These metrics are not enabled yet. These can be enabled using `GrpcServerConfigurer`.

Related PR to add client metric #1021.

PR #1021, opened 2023-12-22
feat: Adds structure for gRPC client metrics instrumentation

DNVindhya:
This PR is one of the initial efforts to support gRFC [A66: OpenTelemetry Metrics](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md) in gRPC-Spring. Even though the gRFC mentions OpenTelmetry, to keep it in sync with existing metrics, this PR as well as future ones will use Micrometer for instrumentation.

This PR adds the structure to collect gRPC metrics for [ClientInterceptor](https://grpc.github.io/grpc-java/javadoc/io/grpc/ClientInterceptor.html). To start with I have only added instrumentation for a single metric i.e `grpc.client.attempt.started`.

PR #1037, opened 2024-01-24
feat: Add gRPC client attempt metrics

DNVindhya:
This PR uses the underlying infrastructure added in #1021 and adds instrumentation for following metrics using Micrometer APIs.

Client per-attempt metrics
- grpc.client.attempt.duration
- grpc.client.attempt.sent_total_compressed_message_size
- grpc.client.attempt.rcvd_total_compressed_message_size

Client per-call metric
- grpc.client.call.duration

These metrics are not enabled yet. These can be enabled using `GrpcGlobalClientInterceptor`.

Related PRs:  
- #1021 
- #1031


DNVindhya:
@ST-DDT Is there a recommended path to add classes that are common between Client and Server metrics?  e.g. `FakeClock` (redefining it for now) will move based on your suggestion.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/metrics/MetricsClientInterceptor.java
```diff
@@ -55,20 +65,22 @@ public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
...
 
-        // TODO(dnvindhya): Collect the actual response/error in the SimpleForwardingClientCall
         return new SimpleForwardingClientCall<ReqT, RespT>(call) {
             @Override
             public void start(Listener<RespT> responseListener, Metadata headers) {
                 delegate().start(
                         new SimpleForwardingClientCallListener<RespT>(responseListener) {
                             @Override
                             public void onClose(Status status, Metadata trailers) {
+                                tracerFactory.callEnded(status);
```
fengli79, line 83: What's the definition of the call duration here?
I see the start is tracked before calling the interceptor chain which the end is tracked before calling the interceptor chain, they are not on pair.
I expect we do either:
- start before calling interceptor chain and end after calling interceptor chain
- start after calling interceptor chain and end before calling interceptor chain
Does it make sense?

DNVindhya: As per `grpc.client.call.duration` [definition](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md#client-per-call-instruments)  end timestamp is calculated before the status is returned to the application. If `traceFactory.callEnded(status)` is called after `super.onClose()` there is no gurantee that `callEnded()` will be called if `onClose` throws any exception [as there is no way to be notified of the exception](https://grpc.github.io/grpc-java/javadoc/io/grpc/ClientCall.Listener.html#onClose(io.grpc.Status,io.grpc.Metadata)).

Regarding the interceptor chain flow, the order of interceptors being invoked follows with the network direction of the call. When we do ClientInterceptors.intercept(A, B), workflow of the call will be 

  Application -> Interceptor B's decorating code for call -> Interceptor A's decorating code for call -> ClientCall (gRPC's code) -> Network

   Network -> ClientCall.Listener (gRPC's code) -> Interceptor A's decorating code for call listener -> interceptor B's decorating code for call listener -> Application
   
 So when we start the call duration timer (e.g in Interceptor B's decorating code for call), it includes time for interceptors added before it and similarly onClose() is called in the reverse direction making it take into account the onClose() of the interceptors added before. So it is symmetric in both ways. 
 Please let me know if that helps.

fengli79: If application has multiple interceptors, based on the SimpleForwardedClientCall, we create a new response listener for each interceptor, each of them have a response listener, however the listeners are not invoked by a chain, and get overwrited each other. Is it the intended? Seems it would break other interceptor's response processing logic.

DNVindhya: IIUC when multiple interceptors are added, response listener callbacks are also invoked as part of the chain, but in the reverse order. This [comment](https://groups.google.com/g/grpc-io/c/g7ovRcBvrcg/m/bGrYQBX5AgAJ) explains `ClientCall.Listener` order and why is it designed in such a way.

fengli79: LGTM.

fengli79 (approved):

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/metrics/MetricsClientInterceptor.java
ST-DDT: > According to `grpc.client.call.duration` [definition](https://github.com/grpc/proposal/blob/master/A66-otel-stats.md?rgh-link-date=2024-01-25T17%3A05%3A52Z#client-per-call-instruments) end timestamp is calculated before the status is returned to the application. If `traceFactory.callEnded(status)` is called after `super.onClose()` there is no gurantee that `callEnded()` will be called if `onClose` throws any exception [as there is no way to be notified of the exception](https://grpc.github.io/grpc-java/javadoc/io/grpc/ClientCall.Listener.html#onClose(io.grpc.Status,io.grpc.Metadata)).

How about using `try-finally`?

DNVindhya: Along with exceptions not getting caught, IIUC `onClose()` is also responsible for closing executors and not guaranteed this will executed in the way we expect. 
Even if use `try-finally`, because of the order in which interceptors are chained and listener callbacks are invoked, invoking it after leads to asymmetric measurement of call duration.

Similar to the example described above,

Application -> Interceptor B's decorating code for call -> Interceptor A's decorating code for call (we start call duration stopwatch here) -> ClientCall (gRPC's code) -> Network

Network -> ClientCall.Listener (gRPC's code) -> Interceptor A's decorating code for call listener (we end call duration stopwatch here) -> interceptor B's decorating code for call listener -> Application

If we invoke it after `super.onClose()`, it will happen after interceptor B's onClose() and that is asymmetric wrt when we started measuring the call duration. 
Please let me know if that makes sense.

ST-DDT (approved):

PR #1039, opened 2024-01-27
Enable gRPC Client attempt and Server metrics

DNVindhya:
This PR enables by adding gRPC client and server metrics defined in the PRs (#1021 , #1031 and #1037) via auto-configuration.

### DIFF
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientMetricAutoConfiguration.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientMetricAutoConfiguration.java
@@ -16,7 +16,6 @@
 
 package net.devh.boot.grpc.client.autoconfigure;
 
-
 import org.springframework.boot.actuate.autoconfigure.metrics.CompositeMeterRegistryAutoConfiguration;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
--- a/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientMetricAutoConfiguration.java
+++ b/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientMetricAutoConfiguration.java
@@ -16,19 +16,25 @@
 
 package net.devh.boot.grpc.client.autoconfigure;
 
+import java.util.function.Supplier;
+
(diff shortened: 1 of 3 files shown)


fengli79 (approved):

DNVindhya:
@ST-DDT  PTAL

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientMetricAutoConfiguration.java
```diff
@@ -40,8 +46,9 @@
 @AutoConfigureAfter(CompositeMeterRegistryAutoConfiguration.class)
 @AutoConfigureBefore(GrpcClientAutoConfiguration.class)
 @ConditionalOnBean(MeterRegistry.class)
-@ConditionalOnClass(MetricCollectingClientInterceptor.class)
+@ConditionalOnClass({MetricCollectingClientInterceptor.class, MetricsClientInterceptor.class})
```
ST-DDT, line 49: Not sure about this condition, because it is basically always true, isn't it?

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientMetricAutoConfiguration.java
```diff
@@ -40,8 +46,9 @@
...
 public class GrpcClientMetricAutoConfiguration {
+    private static final Supplier<Stopwatch> STOPWATCH_SUPPLIER = Stopwatch::createUnstarted;
```
ST-DDT, line 51: At this point in time, this is only used for one method, so you could inline it. 
I don't have any strong feelings regarding that though.

ST-DDT (approved):

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientMetricAutoConfiguration.java
DNVindhya: It is. Updated.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientMetricAutoConfiguration.java
DNVindhya: Done

ST-DDT (dismissed):

ST-DDT (approved):

PR #1041, opened 2024-01-28
Remove the `attempt` base unit for grpc_client_attempt_started metric

fengli79:
Remove the `attempt` base unit for grpc_client_attempt_started metric, to make sure the generated metrics name in the Prometheus is actually grpc_client_attempt_started_total instead of
grpc_client_attempt_started_attempt_total and gets aligned with other languages (C++ for example).


ST-DDT (approved):
What about the server call metrics?
Wouldnt they have the same issue?

fengli79:
> What about the server call metrics? Wouldnt they have the same issue?

Yes, it has the same issue, I'm removing it.

ST-DDT (approved):

PR #1046, opened 2024-02-02
feat: apply grpc service config from consul (#1045)

### LINKED ISSUE #1045: Apply grpc service config from consul
We have [DiscoveryClientNameResolver](https://github.com/grpc-ecosystem/grpc-spring/blob/813288282133ecf43e2f815a847355ac36ef6e21/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java) which allows to specify grpc client target like `discovery:///service-name`. Actual ip:port data is fetched from consul. Port also taken from `gRPC_port` metadata.

I suggest adding `gRPC_service_config` metadata key, similar to `gRPC_port`. Under this key grpc servers may publish [service config](https://github.com/grpc/grpc/blob/master/doc/service_config.md).

This feature already present in standard DNS name resolver ([details](https://github.com/grpc/proposal/blob/master/A2-service-configs-in-dns.md)).

I already implement this in my custom name resolver and it work just fine.

On server side I specify my desired service config:

<details>
  <summary>Server's application.yml</summary>

```yaml
spring:
  cloud:
    consul:
      discovery:
        metadata:
          gRPC_service_config: |
            {
              "loadBalancingConfig": [
                {"round_robin": {}}
              ],
              "methodConfig": [
                {
                  "name": [{}],
                  "retryPolicy": {
                    "maxAttempts": 5,
                    "initialBackoff": "0.05s",
                    "maxBackoff": "1s",
                    "backoffMultiplier": 2,
                    "retryableStatusCodes": [
                      "UNAVAILABLE",
                      "ABORTED",
                      "DATA_LOSS",
                      "INTERNAL",
                      "DEADLINE_EXCEEDED"
                    ]
                  },
                  "timeout": "5s"
                }
              ]
            }
```
</details>

On client side my name resolver parse it and pass to listener:

<details>
  <summary>Code sample</summary>

```java
var result = ResolutionResult.newBuilder().setAddresses(list);
if (!serviceConfig.isEmpty()) {
    try {
        @SuppressWarnings("unchecked")
        Map<String, ?> parsed = gson.fromJson(serviceConfig, Map.class);
        result.setServiceConfig(serviceConfigParser.parseServiceConfig(parsed));
    } catch (JsonSyntaxException e) {
        result.setServiceConfig(
            ConfigOrError.fromError(
                Status.UNKNOWN
                    .withDescription("Failed to parse gRPC_service_config")
                    .withCause(e)
            )
        );
    }
}
savedListener.onResult(result.build());
```
</details>

onyn:
Closes #1045

### DIFF
--- a/grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java
+++ b/grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java
@@ -22,6 +22,7 @@ import static java.util.Objects.requireNonNull;
 import static net.devh.boot.grpc.client.nameresolver.DiscoveryClientResolverFactory.DISCOVERY_INSTANCE_ID_KEY;
 import static net.devh.boot.grpc.client.nameresolver.DiscoveryClientResolverFactory.DISCOVERY_SERVICE_NAME_KEY;
 import static net.devh.boot.grpc.common.util.GrpcUtils.CLOUD_DISCOVERY_METADATA_PORT;
+import static net.devh.boot.grpc.common.util.GrpcUtils.CLOUD_DISCOVERY_METADATA_SERVICE_CONFIG;
 
 import java.net.InetSocketAddress;
 import java.util.List;
@@ -35,6 +36,8 @@ import org.springframework.cloud.client.discovery.DiscoveryClient;
 import org.springframework.util.CollectionUtils;
 
 import com.google.common.collect.Lists;
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
 
 import io.grpc.Attributes;
 import io.grpc.Attributes.Builder;
@@ -58,6 +61,7 @@ public class DiscoveryClientNameResolver extends NameResolver {
     @Deprecated
     private static final String LEGACY_CLOUD_DISCOVERY_METADATA_PORT = "gRPC.port";
     private static final List<ServiceInstance> KEEP_PREVIOUS = null;
+    private static final Gson GSON = new Gson();
 
     private final String name;
     private final DiscoveryClient client;
@@ -65,6 +69,7 @@ public class DiscoveryClientNameResolver extends NameResolver {
     private final Consumer<DiscoveryClientNameResolver> shutdownHook;
     private final SharedResourceHolder.Resource<Executor> executorResource;
     private final boolean usingExecutorResource;
+    private final ServiceConfigParser serviceConfigParser;
 
     // The field must be accessed from syncContext, although the methods on an Listener2 can be called
     // from any thread.
@@ -93,6 +98,7 @@ public class DiscoveryClientNameResolver extends NameResolver {
         this.executor = args.getOffloadExecutor();
         this.usingExecutorResource = this.executor == null;
         this.executorResource = executorResource;
+        this.serviceConfigParser = args.getServiceConfigParser();
     }
 
     /**
@@ -185,6 +191,55 @@ public class DiscoveryClientNameResolver extends NameResolver {
             // TODO: How to handle this case?
             throw new IllegalArgumentException("Failed to parse gRPC port information from: " + instance, e);
         }
+    }
+
+    /**
+     * Extracts and parse gRPC service config from the given service instances.
+     *
+     * @param instances The list of instances to extract the service config from.
+     * @return Parsed gRPC service config or null.
+     */
+    private ConfigOrError resolveServiceConfig(List<ServiceInstance> instances) {
+        final String serviceConfig = getServiceConfig(instances);
+        if (serviceConfig == null) {
+            return null;
+        }
+        log.debug("Found service config for {}", getName());
+        if (log.isTraceEnabled()) {
+            // This is to avoid blowing log into several lines if newlines present in service config string.
+            final String logStr = serviceConfig.replace("\r", "\\r").replace("\n", "\\n");
+            log.trace("Service config for {}: {}", getName(), logStr);
+        }
+        try {
+            @SuppressWarnings("unchecked")
+            Map<String, ?> parsedServiceConfig = GSON.fromJson(serviceConfig, Map.class);
+            return serviceConfigParser.parseServiceConfig(parsedServiceConfig);
+        } catch (JsonSyntaxException e) {
+            return ConfigOrError.fromError(
+                    Status.UNKNOWN
+                            .withDescription("Failed to parse grpc service config")
+                            .withCause(e));
+        }
+    }
+
+    /**
+     * Extracts the gRPC service config string from the given service instances.
+     *
+     * @param instances The list of instances to extract the service config from.
+     * @return The gRPC service config or null.
+     */
+    protected String getServiceConfig(final List<ServiceInstance> instances) {
+        for (final ServiceInstance inst : instances) {
+            final Map<String, String> metadata = inst.getMetadata();
+            if (metadata == null || metadata.isEmpty()) {
+                continue;
+            }
+            final String metaValue = metadata.get(CLOUD_DISCOVERY_METADATA_SERVICE_CONFIG);
+            if (metaValue != null && !metaValue.isEmpty()) {
+                return metaValue;
+            }
+        }
+        return null;
     }
 
     /**
@@ -318,6 +373,7 @@ public class DiscoveryClientNameResolver extends NameResolver {
             log.debug("Ready to update server list for {}", getName());
             this.savedListener.onResult(ResolutionResult.newBuilder()
                     .setAddresses(toTargets(newInstanceList))
+                    .setServiceConfig(resolveServiceConfig(newInstanceList))
                     .build());
             log.info("Done updating server list for {}", getName());
             return newInstanceList;
--- a/docs/en/client/configuration.md
+++ b/docs/en/client/configuration.md
@@ -69,7 +69,7 @@ There are a number of supported schemes, that you can use to determine the targe
 - `discovery` (Prio 6): \
   (Optional) Uses spring-cloud's `DiscoveryClient` to lookup appropriate targets. The connections will be refreshed
   automatically during `HeartbeatEvent`s. Uses the `gRPC_port` metadata to determine the port, otherwise uses the
-  service port. \
+  service port. Uses the `gRPC_service_config` metadata to determine [service config](https://grpc.github.io/grpc/core/md_doc_service_config.html). \
   Example: `discovery:///service-name`
 - `self` (Prio 0): \
   The self address or scheme is a keyword that is available, if you also use `grpc-server-spring-boot-starter` and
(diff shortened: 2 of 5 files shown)


ST-DDT (changes requested):
Could you please replace `var` with actual types?

grpc-client-spring-boot-autoconfigure/src/test/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolverTest.java
```diff
@@ -0,0 +1,119 @@
...
+                      },
+                      "timeout": "5s"
+                    }
+                  ]
+                }
+                """;
+        TestableListener listener = resolveServiceAndVerify("test1", validServiceConfig);
+        NameResolver.ConfigOrError serviceConf = listener.getResult().getServiceConfig();
+        assertThat(serviceConf).isNotNull();
+        assertThat(serviceConf.getConfig()).isNotNull();
```
ST-DDT, line 85: Please check at least one value from the config.

grpc-client-spring-boot-autoconfigure/src/test/java/net/devh/boot/grpc/client/nameresolver/TestableListener.java
```diff
@@ -0,0 +1,46 @@
...
+@Getter
+public class TestableListener extends NameResolver.Listener2 {
+
+    private NameResolver.ResolutionResult result;
+
+    private Status error;
+
+    private boolean resultWasSet = false;
+
+    private boolean errorWasSet = false;
```
ST-DDT, line 32:
```suggestion
    private NameResolver.ResolutionResult result;
    private Status error;
    private boolean resultWasSet = false;
    private boolean errorWasSet = false;
```

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java
```diff
@@ -314,11 +341,29 @@ private List<ServiceInstance> resolveInternal() {
...
+                    Map<String, ?> parsedServiceConfig = GSON.fromJson(serviceConfig, Map.class);
+                    result.setServiceConfig(serviceConfigParser.parseServiceConfig(parsedServiceConfig));
+                } catch (JsonSyntaxException e) {
+                    result.setServiceConfig(
+                            ConfigOrError.fromError(
+                                    Status.UNKNOWN
+                                            .withDescription("Failed to parse grpc service config")
+                                            .withCause(e)));
+                }
+            }
```
ST-DDT, line 362: Could you please move this block to a separate method?
Maybe let it return `ConfigOrError`. I wished there was a method for that in the grpc api/core library.

Also the DNSNameresolver does a lot more with the configs, not sure which parts we should copy from that:
https://github.com/grpc/grpc-java/blob/3202370684ec74a8e986a116b3e1476ff0a3fedc/core/src/main/java/io/grpc/internal/DnsNameResolver.java#L489

grpc-client-spring-boot-autoconfigure/src/test/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolverTest.java
onyn: I tried. But `getConfig()` returns object of type `Object`. Real type of this object is `io.grpc.internal.ManagedChannelServiceConfig` which have package level visibility. Value test can be done only by using reflection or creating proxy class in `io.grpc.internal` package which opens `ManagedChannelServiceConfig` for test. But this makes test fragile. I can do it if you ok with this.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java
onyn: Ok, will move it.

DnsNameResolver uses array of service configs with rules to which service which config should be applied ([more on this](https://github.com/grpc/proposal/blob/master/A2-service-configs-in-dns.md#canarying-changes)). I choose not to bring in this complication. But if you want, I can implement this exactly like in DnsNameResolver.

grpc-client-spring-boot-autoconfigure/src/test/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolverTest.java
ST-DDT: If it is that deeply hidden, then we dont need additional test for it.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java
ST-DDT: It looks like they already put quite some though into it. I would like to get feedback from the other grpc developers on this.
@DNVindhya Do you have a contact that could enlighten us or help us getting an API method for String->ServiceConfig?

DNVindhya: Opening a grpc-java issue for this will be apt.

ST-DDT:
Please resolve the merge conflicts.

grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java
ST-DDT: @onyn any updates one this? Do you have a link for the api request?

onyn:
Done. Two questions remains:

1. Should we support [canarying changes](https://github.com/grpc/proposal/blob/master/A2-service-configs-in-dns.md#canarying-changes) feature of DNS resolver? This may be supported in the future, using separate metadata key in case of demand.
2. Currently GRPC doesn't expose methods for parsing json string into `ConfigOrError` object. Shoud we wait for it?

ST-DDT:
Thanks for your fast response.

Lets wait a bit on how upstream decides on the API topic.

onyn:
[That](https://github.com/grpc/grpc-java/issues/11008#issuecomment-2010949346) was expected.

So, just one question left:

1. Should we support [canarying changes](https://github.com/grpc/proposal/blob/master/A2-service-configs-in-dns.md#canarying-changes) feature of DNS resolver? This may be supported in the future, using separate metadata key in case of demand.
2. ~~Currently GRPC doesn't expose methods for parsing json string into `ConfigOrError` object. Shoud we wait for it?~~

ST-DDT:
> 1. Should we support [canarying changes](https://github.com/grpc/proposal/blob/master/A2-service-configs-in-dns.md?rgh-link-date=2024-03-25T09%3A44%3A59Z#canarying-changes) feature of DNS resolver?

No, just keep it simple. If grpc adds an official API, we can adopt it then.

onyn:
Rebased and squashed.

grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java
```diff
@@ -314,11 +369,13 @@ private List<ServiceInstance> resolveInternal() {
...
+            final ResolutionResult.Builder result = ResolutionResult.newBuilder()
+                    .setAddresses(toTargets(newInstanceList))
+                    .setServiceConfig(resolveServiceConfig(newInstanceList));
+
             // Set new servers
             log.debug("Ready to update server list for {}", getName());
-            this.savedListener.onResult(ResolutionResult.newBuilder()
-                    .setAddresses(toTargets(newInstanceList))
-                    .build());
+            this.savedListener.onResult(result.build());
```
ST-DDT, line 378: IMO you can directly build the result in line 372 instead of delaying it till here. There aren't any intermediate calls to the builder so we can improve the readability by building early.

grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java
```diff
@@ -187,6 +193,55 @@ protected int getGrpcPort(final ServiceInstance instance) {
...
+    }
+
+    /**
+     * Extracts the gRPC service config string from the given service instances.
+     *
+     * @param instances The list of instances to extract the service config from.
+     * @return The gRPC service config or null.
+     */
+    protected String getServiceConfig(final List<ServiceInstance> instances) {
+        for (final ServiceInstance inst : instances) {
```
ST-DDT, line 232: Should we shuffle them to add some randomness to this?

onyn: Better to refuse any config if there's differences in any single service from others. And log it with "warn" level to make the problem visible. There's no sense to prefer one config over another, because all service instances are equivalent from grpc's point of view. We should force service operators to configure their instances with same service config.

ST-DDT: The thing is that when the config changes, it will only slowly propagate to all new instances.
It would then cause these warnings.
Not sure about selecting a random one though. Maybe a sticky random (keep the config while it is provided by at least one instance).

onyn: That makes sense. Anyway, I see no benefits in randomness (including sticky one). We can't do anything in case of human error while configuring service config. And predictable approach works better in case of rolling update.

grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/nameresolver/DiscoveryClientNameResolver.java
```diff
@@ -318,7 +373,7 @@ private List<ServiceInstance> resolveInternal() {
             log.debug("Ready to update server list for {}", getName());
             this.savedListener.onResult(ResolutionResult.newBuilder()
                     .setAddresses(toTargets(newInstanceList))
-                    .build());
+                    .setServiceConfig(resolveServiceConfig(newInstanceList)).build());
```
ST-DDT, line 376:
```suggestion
                    .setServiceConfig(resolveServiceConfig(newInstanceList))
                    .build());
```

ST-DDT (approved):
Looks good to me.

onyn:
Rebased & squashed.

ST-DDT (approved):

yidongnan (approved):

PR #1072, opened 2024-03-18
compileOnly for grpc-inprocess/implementation for grpc-netty-shaded

panchenko:
* The `grpc-inprocess` module exists mostly for testing purposes, we don't need it at runtime.
* The classes inside `grpc-netty-shaded` are repackaged, users are not supposed to access those internal packages, so implementation is enough

### DIFF
--- a/grpc-client-spring-boot-starter/build.gradle
+++ b/grpc-client-spring-boot-starter/build.gradle
@@ -26,8 +26,8 @@ dependencies {
     //  this means it can be used implicitly without specifying the dependency (unless you wish to override with a different version, or exclude)
     optionalSupportApi 'io.grpc:grpc-netty'
     optionalSupportApi 'io.netty:netty-transport-native-epoll'
-    api 'io.grpc:grpc-inprocess'
-    api 'io.grpc:grpc-netty-shaded'
+    compileOnly 'io.grpc:grpc-inprocess'
+    implementation 'io.grpc:grpc-netty-shaded'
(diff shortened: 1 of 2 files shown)


shalk:
thought this is right. But may be user have to always add grpc-inprocess. Because [GrpcClientAutoConfiguration](https://github.com/grpc-ecosystem/grpc-spring/blob/master/grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java) have add InProcessChannelFactory for  each GrpcChannelFactory case

https://github.com/grpc-ecosystem/grpc-spring/blob/master/grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java#L152

https://github.com/grpc-ecosystem/grpc-spring/blob/master/grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java#L170

https://github.com/grpc-ecosystem/grpc-spring/blob/master/grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java#L170

panchenko:
@shalk Thanks for the feedback, I will check the client more and might be split the change into multiple PRs to simplify the discussion.
Actually my main interest is the server, it works with grpc-inprocess excluded via gradle.

PR #1081, opened 2024-03-27
Add instrumentation tags to grpc metrics

stanley-cheung:
Add two tags to the A66 metrics exported by the MetricsStreamTracers

```
  'instrumentation_source': 'grpc-spring'
  'instrumentation_version': {current project version}
```


fengli79 (approved):

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Versions.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Class that contains version strings
+ */
+public final class Versions {
+
+    /**
+     * A constant that defines, the current version of the library.
+     */
+    public static final String PROJECT_VERSION = "v3.0.0.RELEASE";
```
fengli79, line 27: We should not hardcode this. It needs come from the build process.

fengli79 (changes requested):

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Versions.java
stanley-cheung: Fixed

stanley-cheung: I end up taking a slightly different approach. Instead of doing a string replace, which will _modify the source file_, which seems undesirable, I now _generate_ the `Versions.java` file as part of the `build.gradle` build step. This seems cleaner, because that generated file can later be cleaned up by `./gradlew clean`.

stanley-cheung:
I can't quite figure out how to fix the current gradle build error.

```
$ ./gradlew publishToMavenLocal


> Task :grpc-common-spring-boot:sourcesJar FAILED

FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':grpc-common-spring-boot:sourcesJar' (type 'Jar').
  - Gradle detected a problem with the following location: '/usr/local/google/home/stanleycheung/grpc-spring/grpc-common-spring-boot/src/main/java'.
    
    Reason: Task ':grpc-common-spring-boot:sourcesJar' uses this output of task ':createVersionsFile' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':createVersionsFile' as an input of ':grpc-common-spring-boot:sourcesJar'.
      2. Declare an explicit dependency on ':createVersionsFile' from ':grpc-common-spring-boot:sourcesJar' using Task#dependsOn.
      3. Declare an explicit dependency on ':createVersionsFile' from ':grpc-common-spring-boot:sourcesJar' using Task#mustRunAfter.
```

stanley-cheung:
Some more work to be done...

grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/metrics/MetricsClientStreamTracers.java
```diff
@@ -127,7 +128,10 @@ public void streamClosed(Status status) {
 
         void recordFinishedAttempt() {
             Tags attemptMetricTags =
-                    Tags.of("grpc.method", fullMethodName, "grpc.status", statusCode.toString());
+                    Tags.of("grpc.method", fullMethodName,
+                            "grpc.status", statusCode.toString(),
+                            "instrumentation_source", Constants.INSTRUMENTATION_SOURCE_TAG_VALUE,
```
fengli79, line 133: Also, the label keys can be encapsulated in this file instead of define in the Constants.java.

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Constants.java.template
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * A constant that defines the current version of the library.
+     */
+    public static final String PROJECT_VERSION = "@versionStringPlaceholder@";
+
+
+    /**
+     * A constant that defines the instrumentation_source metric tag value
+     */
+    public static final String INSTRUMENTATION_SOURCE_TAG_VALUE = "grpc-spring";
```
fengli79, line 33: s/INSTRUMENTATION_SOURCE_TAG_VALUE/LIBRARY_NAME

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Constants.java.template
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Class that contains shared constants
+ */
+public final class Constants {
+
+    /**
+     * A constant that defines the current version of the library.
+     */
+    public static final String PROJECT_VERSION = "@versionStringPlaceholder@";
```
ST-DDT, line 27: Instead of project build tools you could also use `Constants.class.getPackage().getImplementationVersion()` (if the related properties for the jar goal are configured)

grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/metrics/MetricsClientStreamTracers.java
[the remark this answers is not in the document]
stanley-cheung: Fixed.

grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/metrics/MetricsClientStreamTracers.java
```diff
@@ -127,7 +128,10 @@ public void streamClosed(Status status) {
...
+                            "instrumentation_version", Constants.PROJECT_VERSION);
```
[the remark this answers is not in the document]
stanley-cheung: Fixed

grpc-client-spring-boot-starter/src/main/java/net/devh/boot/grpc/client/metrics/MetricsClientStreamTracers.java
stanley-cheung: Done.

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Constants.java.template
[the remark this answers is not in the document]
stanley-cheung: Fixed

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Constants.java.template
stanley-cheung: Fixed

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Constants.java.template
stanley-cheung: Will look into that, probably in a subsequent PR. Thanks for the suggestions.

stanley-cheung: Actually that works. I removed the build generation step. Thanks for the suggestion again!

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Constants.java
```diff
@@ -0,0 +1,38 @@
...
+ * Class that contains shared constants
+ */
+public final class Constants {
+
+    /**
+     * A constant that defines the current version of the library.
+     */
+    public static final String VERSION = "v" + Constants.class.getPackage().getImplementationVersion();
+
+
```
ST-DDT, line 29:
```suggestion

```

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Constants.java
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * A constant that defines the library name that can be used as metric tags
+     */
+    public static final String LIBRARY_NAME = "grpc-spring";
+
+
```
ST-DDT, line 35:
```suggestion

```

grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/Constants.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.devh.boot.grpc.common.util;
+
+/**
+ * Class that contains shared constants
```
ST-DDT, line 20:
```suggestion
 * Class that contains shared constants.
```

stanley-cheung:
> > Add two tags to the A66 metrics exported by the MetricsStreamTracers
> 
> I haven't found a reason why these tags should be included. Is it part of the spec?

Yes this is part of an internal spec that will help us distinguish where the metrics originate.

ST-DDT (approved):
I'm not sure whether the tags should be included for everyone.
But implementation wise this looks good to me.
(Spring has a bean for adding tags to metrics).

DNVindhya (approved):
LGTM

fengli79 (approved):

Issue #1088, opened 2024-04-09
GrpcExceptionListener handle exceptions outside of observation scope


hnvas:
**The bug**

The GrpcAdviceAutoConfiguration class initialises the global exception handler with higher precedence than the GrpcServerMicrometerTraceAutoConfiguration class, causing the GrpcExceptionListener to handle exceptions outside of observation scope and making it impossible for the exception handlers declared in @GrpcAdvice annotated classes to access the current tracing context.

**Stacktrace and logs**

```
[2024-04-09 14:31:42.548] - [grpc-default-executor-0] - [INFO] - [com.hnvas.grpctracingdemo.ExampleApiImpl->sayHelloError:26] - [traceId=66157b7e32dfc6c022c9a26f05d49218, spanId=22c9a26f05d49218]: Received a request on sayHelloError
[2024-04-09 14:31:42.549] - [grpc-default-executor-0] - [ERROR] - [com.hnvas.grpctracingdemo.GrpcExceptionAdvice->handleIllegalStateException:15] - [traceId=, spanId=]: This should not happen: null
java.lang.IllegalStateException: This should not happen
	at com.hnvas.grpctracingdemo.ExampleApiImpl.sayHelloError(ExampleApiImpl.java:27)
	at com.hnvas.grpctracingdemo.grpc.ExampleApiGrpc$MethodHandlers.invoke(ExampleApiGrpc.java:276)
	at io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:182)
	at io.grpc.PartialForwardingServerCallListener.onHalfClose(PartialForwardingServerCallListener.java:35)
	at io.grpc.ForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:23)
	at io.grpc.ForwardingServerCallListener$SimpleForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:40)
	at io.micrometer.core.instrument.binder.grpc.ObservationGrpcServerCallListener.lambda$onHalfClose$1(ObservationGrpcServerCallListener.java:47)
	at io.micrometer.observation.Observation.scoped(Observation.java:709)
	at io.micrometer.core.instrument.binder.grpc.ObservationGrpcServerCallListener.onHalfClose(ObservationGrpcServerCallListener.java:47)
[...]

PR #1134, opened 2024-08-15
fix: changes grpc server exception handler order

### LINKED ISSUE #789: Server metrics are tagged with the wrong status codes
**The context**

I'd like to get the metric values for `grpc.server.processing.duration` and distinguish different status codes, e.g. OK, NOT_FOUND, INVALID_ARGUMENT, UNKNOWN

**The bug**

If my gRPC service is called three times, and it responds twice with an OK status and once with with an NOT_FOUND status, I'd expect the tags that are captured for the metrics `grpc.server.processing.duration` to be "OK" and "NOT_FOUND".
But instead, the "NOT_FOUND" error seems to be reported as an "UNKNOWN" error.

**Stacktrace and logs**

I don't know

**Steps to Reproduce**

I created a very simple [example project](https://github.com/harti2006/grpc-metrics-demo), with one gRPC service and one GrpcAdvice that translates my "business" exception into a NOT_FOUND status.

Please check out the integration test, that shows how to reproduce the issue: https://github.com/harti2006/grpc-metrics-demo/blob/trunk/src/test/java/com/github/harti2006/grpc/metrics/GrpcMetricsDemoApplicationTests.java

Which versions do you use?
* Spring (boot): 2.7.5
* grpc-java: 1.51.0
* grpc-spring-boot-starter: 2.14.0.RELEASE
* java: 11

**Additional context**

* It used to work fine with the previous release `2.13.1.RELEASE`. When downgrading to this version, the above mentioned integration test works fine

s4kas:
This PR adds the `GrpcExceptionInterceptor` bean after the `ObservationGrpcServerInterceptor` and the `MetricCollectingServerInterceptor` and before the `ExceptionTranslatingServerInterceptor`.
This is necessary in order to allow the later interceptor to process the error response after the exception handler is executed, more precisely to add the returned `status code` to the context/meter.

Im not entirely sure but i believe this was introduced as an unintentional consequence of the added support for [Order](https://github.com/grpc-ecosystem/grpc-spring/pull/737/files#diff-96de4406135a56a86c3443cb117b6fa3eb9b73213e3753a456d0e181f1860930) because the [GrpcAdviceAutoConfiguration](https://github.com/grpc-ecosystem/grpc-spring/blob/a18aeeb25e60114ea773a621bfdc44902513ed07/grpc-server-spring-boot-starter/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcAdviceAutoConfiguration.java#L71) already had the `@Order` with the [0 precedence](https://github.com/grpc-ecosystem/grpc-spring/blob/a18aeeb25e60114ea773a621bfdc44902513ed07/grpc-common-spring-boot/src/main/java/net/devh/boot/grpc/common/util/InterceptorOrder.java#L48) it was just not doing anything.

fixes #789 #1088

### DIFF
--- a/tests/src/test/java/net/devh/boot/grpc/test/advice/GrpcAdviceWithMetricsTest.java
+++ b/tests/src/test/java/net/devh/boot/grpc/test/advice/GrpcAdviceWithMetricsTest.java
@@ -1,0 +1,112 @@
+
+package net.devh.boot.grpc.test.advice;
+
+import static io.grpc.Status.INVALID_ARGUMENT;
+import static io.micrometer.core.instrument.binder.grpc.GrpcObservationDocumentation.LowCardinalityKeyNames.STATUS_CODE;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.ImportAutoConfiguration;
+import org.springframework.boot.test.autoconfigure.actuate.observability.AutoConfigureObservability;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.test.annotation.DirtiesContext;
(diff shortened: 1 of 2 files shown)


tests/src/test/java/net/devh/boot/grpc/test/advice/GrpcAdviceWithMetricsTest.java
```diff
@@ -0,0 +1,112 @@
...
+                return INVALID_ARGUMENT.withCause(e);
+            }
+        }
+
+        @GrpcService
+        public static class TestGrpcAdviceService extends TestServiceGrpc.TestServiceImplBase {
+            @Override
+            public void error(Empty request, StreamObserver<Empty> responseObserver) {
+                throw new RuntimeException("a simulated error");
+            }
```
ST-DDT, line 109: Please also add a streaming variant of this test.

Also I'm thinking of whether we should have an additional test for `responseObserver.onError(exception)` in addition to just throwing inside the method.

